PR #21, opened 2017-02-24
Switch from TestNG to JUnit

livthomas:


livthomas:
Already done in pull-request #20.

PR #24, opened 2017-03-09
#LMR-73 project and org. scoping in ConfigurationFacade

marvec:


war/src/main/java/io/lumeer/engine/controller/ConfigurationFacade.java
```diff
@@ -103,11 +106,11 @@ public Boolean getDataStorageUseSsl() {
     * @return Optional Integer value of the given key
     */
    public Optional<Integer> getConfigurationInteger(final String key) {
-      Optional<String> value = getConfigurationString(key);
-      if (value.isPresent()) {
-         return Optional.of(Integer.parseInt(value.get()));
+      try {
+         return getConfigurationString(key).map(Integer::parseInt);
+      } catch (NumberFormatException nfe) {
+         return Optional.empty();
```
livthomas: Is this really a good idea? Do not we want to at least trace log in such situations?

war/src/main/java/io/lumeer/engine/controller/ConfigurationFacade.java
```diff
@@ -276,50 +295,86 @@ public void setUserConfigurationInteger(final String key, final int value) {
...
     *
     * @param key
     *       the name of key
     * @param document
     *       the DataDocument of the given key
     */
-   public void setTeamConfigurationDocument(final String key, final DataDocument document) {
-      setConfiguration(false, key, document);
+   public void setOrganisationConfigurationDocument(final String key, final DataDocument document) {
+      setConfiguration(ConfigurationLevel.ORGANISATION, key, document);
```
livthomas: I can understand why we have different getters for `String` and `Integer` values. But why do we have setters for different types when we can use just `Object`?

war/src/main/java/io/lumeer/engine/controller/ConfigurationFacade.java
```diff
@@ -329,52 +384,81 @@ public void resetUserConfiguration() {
...
    }
 
    /**
-    * Resets currently logged team configuration. The whole team configuration entry will not be deleted from the system collection, just the config field!
+    * Resets currently logged organisation configuration. The whole organisation configuration entry will not be deleted from the system collection, just the config field!
     */
-   public void resetTeamConfiguration() {
-      resetConfiguration(false, null);
+   public void resetOrganisationConfiguration() {
+      resetConfiguration(ConfigurationLevel.ORGANISATION, null);
```
livthomas: I do not like this approach. Why do we reset configuration by setting `null`? We can accidentally reset the configuration when `null` value is passed without us noticing it.

war/src/main/java/io/lumeer/engine/util/Resources.java
```diff
@@ -56,7 +56,11 @@
    @Produces
    @Dependent
    public Logger produceLog(InjectionPoint injectionPoint) {
-      return Logger.getLogger(injectionPoint.getMember().getDeclaringClass().getName());
+      return produceLog(injectionPoint.getMember().getDeclaringClass().getName());
+   }
+
+   public static Logger produceLog(final String className) {
+      return Logger.getLogger(className);
```
livthomas: What is this change good for?

livthomas: What is this change good for?

war/src/main/java/io/lumeer/engine/controller/ConfigurationFacade.java
livthomas: (There are more similar fragments of code.)

war/src/main/java/io/lumeer/engine/util/Resources.java
marvec: Logger cannot be directly injected into serializable CDI beans (app., session and conversation scoped). So this is a workaround to ease its creation as some contributors used println instead.

PR #26, opened 2017-03-11
Completed LMR-75 and LMR-83

marvec:


marvec:
I will dare to merge it now, so everybody can continue with development on this update tomorrow. I won't be available for most of the day as I travel.

PR #27, opened 2017-03-13
Redesigned CollectionMetadataFacade

alicak:


alicak:
There is a problem with method renameAttribute in CollectionMetadataFacade.

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -176,236 +184,118 @@ public String createInternalName(String originalCollectionName) throws UserColle
...
-         return false;
-      }
-
-      List<DataDocument> newAttributeInfo = dataStorage.run(queryCollectionAttributeInfo(collectionName, newName));
-
-      // check if the attribute with new name already exists in the collection
-      if (!newAttributeInfo.isEmpty()) {
+   public void renameAttribute(String collectionName, String oldName, String newName) throws AttributeAlreadyExistsException {
+      if (getAttributesNames(collectionName).contains(newName)) {
          throw new AttributeAlreadyExistsException(ErrorMessageBuilder.attributeAlreadyExistsString(newName, collectionName));
```
livthomas: +1 on moving this check at the beginning of the method and putting the rest of the code outside the `if` statement. It should always be done this way in order to avoid unnecessary nesting.

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -532,80 +424,52 @@ public String getInternalCollectionName(String originalCollectionName) throws Us
...
-    * Sets collection lock to new value
+    * Sets the time of last collection usage to current time
     *
     * @param collectionName
     *       internal collection name
-    * @param newTime
-    *       String representation of the time of the last update of collection lock
-    * @return true if set was successful, false if time format is wrong or metadata collection does not exist
     */
-   public boolean setCollectionLockTime(String collectionName, String newTime) {
```
livthomas: Why are `getCollectionLockTime` and  `setCollectionLockTime` methods being removed?

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -1176,4 +821,9 @@ private void initConstraintManager(ConstraintManager constraintManager) {
    private String getCurrentUser() {
       return userFacade.getUserEmail();
    }
+
+   // returns string "parent.child"
+   private String nestedAttributeName(String parent, String child) {
```
livthomas: Can be `static`.

war/src/main/java/io/lumeer/engine/rest/dao/CollectionMetadata.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+package io.lumeer.engine.rest.dao;
+
+import io.lumeer.engine.api.LumeerConst;
+import io.lumeer.engine.api.data.DataDocument;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.xml.crypto.Data;
```
livthomas: This import is not needed.

war/src/test/java/io/lumeer/engine/controller/CollectionFacadeIntegrationTest.java
```diff
@@ -100,9 +102,10 @@ public void testReadCollectionMetadata() throws Exception {
       String name = "attribute 1";
       collectionMetadataFacade.addOrIncrementAttribute(collection, name);
 
-      List<DataDocument> metadata = collectionFacade.readCollectionMetadata(collection);
+      CollectionMetadata metadata = collectionFacade.readCollectionMetadata(collection);
 
-      assertThat(metadata).hasSize(4); // 4 documents: attribute, name, lock, rights
+      // TODO
+      assertThat(metadata); // 4 documents: attribute, name, lock, rights
```
livthomas: Why has this assert been changed?

war/src/test/java/io/lumeer/engine/controller/CollectionMetadataFacadeIntegrationTest.java
```diff
@@ -148,43 +150,19 @@ public void testRenameCollectionAttribute() throws Exception {
...
-
-      // we try to retype to invalid type
-      String newType = "hello";
-      retype = collectionMetadataFacade.retypeCollectionAttribute(collection, name, newType);
-      assertThat(collectionMetadataFacade.getAttributeType(collection, name)).isEqualTo(oldType);
-      assertThat(retype).isFalse();
-
-      // we try to retype non existing attribute
-      assertThat(collectionMetadataFacade.retypeCollectionAttribute(collection, "attribute2", LumeerConst.Collection.COLLECTION_ATTRIBUTE_TYPE_STRING)).isFalse();
+      // TODO
```
livthomas: What should be here?

war/src/test/java/io/lumeer/engine/controller/CollectionMetadataFacadeIntegrationTest.java
```diff
@@ -231,39 +209,21 @@ public void testGetInternalCollectionName() throws Exception {
...
+
+      assertThat(metadata.getName()).isEqualTo(COLLECTION_CREATE_INITIAL_METADATA);
+      assertThat(metadata.getInternalName()).isEqualTo(collection);
+      assertThat(metadata.getProjectId()).isEqualTo(projectFacade.getCurrentProjectId());
+      assertThat(metadata.getAttributes()).isEmpty();
+      assertThat(metadata.getLastTimeUsed()).isNotEmpty();
+      assertThat(metadata.getRecentlyUsedDocumentIds()).isEmpty();
+      assertThat(metadata.getCreateDate()).isNotEmpty();
+      assertThat(metadata.getCreator()).isEqualTo(userFacade.getUserEmail());
+      assertThat(metadata.getCustomMetadata()).isEmpty();
```
livthomas: [Soft assertions](http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#soft-assertions) would be better here.

war/src/test/java/io/lumeer/engine/controller/CollectionMetadataFacadeIntegrationTest.java
```diff
@@ -523,7 +344,7 @@ public void testIsUserCollection() {
    //
    //      String name = "attribute 1";
    //      collectionMetadataFacade.addOrIncrementAttribute(collection, name);
-   //      collectionMetadataFacade.retypeCollectionAttribute(collection, name, collectionMetadataFacade.COLLECTION_ATTRIBUTE_TYPE_INT);
+   //      collectionMetadataFacade.retypeCollectionAttribute(collection, name, collectionMetadataFacade.ATTRIBUTE_TYPE_INT);
```
livthomas: Why is this test commented? It should either be fixed, removed or marked as ignored with a good explanation.

war/src/test/java/io/lumeer/engine/rest/CollectionServiceIntegrationTest.java
```diff
@@ -461,33 +464,33 @@ public void testUpdateAccessRights() throws Exception {
...
-      Response response2 = client.target(TARGET_URI).path(PATH_PREFIX + COLLECTION_SET_AND_READ_ATTRIBUTE_TYPE + "/attributes/" + attributeName + "/types").request(MediaType.APPLICATION_JSON).buildGet().invoke();
-      String attributeType = response2.readEntity(String.class);
-      assertThat(response2.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
-      assertThat(attributeType).isEqualTo(newType);
-      response2.close();
-
-      client.close();
-   }
+//   @Test
+//   public void testSetAndReadAttributeType() throws Exception {
```
livthomas: The same applies here. If it is no longer needed, this test method should be removed.

mongodb-storage/src/main/java/io/lumeer/mongodb/MongoDbStorageDialect.java
```diff
@@ -39,22 +39,7 @@
...
-                        new DataDocument(LumeerConst.Collection.META_TYPE_KEY, LumeerConst.Collection.COLLECTION_ATTRIBUTES_META_TYPE_VALUE)
-                              .append(LumeerConst.Collection.COLLECTION_ATTRIBUTE_NAME_KEY, attributeName)
-                              .append(LumeerConst.Collection.COLLECTION_ATTRIBUTE_TYPE_KEY, LumeerConst.Collection.COLLECTION_ATTRIBUTE_TYPE_STRING)
-                              .append(LumeerConst.Collection.COLLECTION_ATTRIBUTE_CONSTRAINTS_KEY, new ArrayList<String>())
-                  )
-                        .append("$inc",
-                              new DataDocument(LumeerConst.Collection.COLLECTION_ATTRIBUTE_COUNT_KEY, 1)))
-            .append("new", true)
-            .append("upsert", true);
+      return new DataDocument(); // TODO based on new query in CollectionMetadataFacade
```
alicak: I am planning to move all queries from CollectionMetadataFacade to this class after CollectionMetadataFacade is finished.

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
alicak: They were never used and I thought that they were actually replaced by `setLastTimeUsedNow` and `getLastTimeUsed`.

war/src/main/java/io/lumeer/engine/controller/CollectionFacade.java
```diff
@@ -113,127 +115,86 @@ public void init() {
...
     * @throws UserCollectionAlreadyExistsException
     *       when collection with given user name already exists
-    * @throws CollectionNotFoundException
-    *       when newly created metadata collection for the collection was not found and initial metadata could not be created
     */
-   public String createCollection(final String collectionOriginalName) throws CollectionAlreadyExistsException, UserCollectionAlreadyExistsException, CollectionNotFoundException {
+   public String createCollection(final String collectionOriginalName) throws UserCollectionAlreadyExistsException {
       String internalCollectionName = collectionMetadataFacade.createInternalName(collectionOriginalName);
 
-      if (!dataStorage.hasCollection(internalCollectionName)) {
```
alicak: The check is performed inside `createInternalName`.

war/src/main/java/io/lumeer/engine/controller/CollectionFacade.java
```diff
@@ -113,127 +115,86 @@ public void init() {
...
+    * @return an object with collection metadata
     */
-   public List<DataDocument> readCollectionMetadata(final String collectionName) throws CollectionNotFoundException {
-      if (dataStorage.hasCollection(collectionName)) {
-         return dataStorage.search(collectionMetadataFacade.collectionMetadataCollectionName(collectionName), null, null, 0, 0);
-      } else {
-         throw new CollectionNotFoundException(ErrorMessageBuilder.collectionNotFoundString(collectionName));
-      }
+   public CollectionMetadata readCollectionMetadata(final String collectionName)  {
+      return collectionMetadataFacade.getCollectionMetadata(collectionName);
```
alicak: It is a relict from initial design of facades :-) I'll remove it.

war/src/test/java/io/lumeer/engine/controller/CollectionFacadeIntegrationTest.java
alicak: Because before that we were getting result as DataDocument and now it is CollectionMetadataObject. However, I am going to remove the method that is tested, so the test will be removed too.

livthomas:
The following code seems to work for me:
```java
DataDocument renameQuery = new DataDocument()
                  .append("findAndModify", METADATA_COLLECTION)
                  .append("query", new DataDocument(LumeerConst.Collection.INTERNAL_NAME_KEY, collectionName))
                  .append("update", new DataDocument()
                        .append("$rename", new DataDocument(
                              nestedAttributeName(LumeerConst.Collection.ATTRIBUTES_KEY, oldName),
                              nestedAttributeName(LumeerConst.Collection.ATTRIBUTES_KEY, newName))));
            dataStorage.run(renameQuery);
```

war/src/test/java/io/lumeer/engine/rest/CollectionServiceIntegrationTest.java
```diff
@@ -145,6 +145,7 @@ public void testGetAllCollections() throws Exception {
       client.close();
    }
 
+   @Ignore // we ignore that test because we don't use cache anymore
```
livthomas: This comment can be added as a parameter of `@Ignore` annotation.

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -245,33 +247,35 @@ public void createInitialMetadata(String internalCollectionName, String original
...
-            METADATA_COLLECTION,
-            new DataDocument(
-                  nestedAttributeName(LumeerConst.Collection.ATTRIBUTES_KEY, newName),
-                  attribute),
-            documentId);
+
+      // still does not pass the test, even with $rename
+      DataDocument renameQuery = new DataDocument()
+            .append("findAndModify", METADATA_COLLECTION)
+            .append("query", new DataDocument(LumeerConst.Document.ID, documentId))
```
livthomas: See the code snippet in my previous comment. The following line should be here:
```java
.append("query", new DataDocument(LumeerConst.Collection.INTERNAL_NAME_KEY, collectionName))
```

And the previous getting of metadata document is not necessary.

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -677,7 +682,11 @@ public void dropAttributeConstraint(String collectionName, String attributeName,
     * @return true if user can read the collection
     */
    public boolean checkCollectionForRead(String collectionName, String user) {
-      return securityFacade.checkForRead(getCollectionMetadataDocument(collectionName), user);
+      try {
+         return securityFacade.checkForRead(getCollectionMetadataDocument(collectionName), user);
+      } catch (CollectionMetadataDocumentNotFoundException e) {
+         return true; // if metadata is not found, we allow access
```
livthomas: Should not there always be a metadata document? Or when can this happen?

alicak: I thought it would be better to check existence of the document, but if it does not exist, something bad is happening which should never happen.

livthomas:
However, there were some changes on the `devel` branch in the meantime. So in order to merge this, we need to resolve the conflicts.

api/src/main/java/io/lumeer/engine/api/data/DataDocument.java
```diff
@@ -261,7 +261,13 @@ public boolean getBoolean(final String key, final boolean defaultValue) {
...
     */
    public Date getDate(final String key) {
-      return (Date) getObject(key);
+      Long date = null;
+      try {
+         date = (Long) getObject(key);
+      } catch (ClassCastException e) {
+         return (Date) getObject(key);
+      }
+      return new Date(date);
```
livthomas: I am afraid this will fail if the value of `date` is `null`. I would suggest adding a check if the value of `date` is not `null` on the last line and only in that case pass the value to the constructor.

PR #32, opened 2017-03-26
LMR-79 database stats

marvec:


api/src/main/java/io/lumeer/engine/api/data/DataStorageStats.java
```diff
@@ -0,0 +1,155 @@
...
+
+   private String databaseName;
+
+   private long collections;
+
+   private long documents;
+
+   private long dataSize;
+
+   private long storageSize;
```
livthomas: It would be useful to add a JavaDoc explaining these attributes. Especially the difference between `dataSize` and `storageSize` is not obvious unless you read MongoDB documentation.

PR #33, opened 2017-03-27
Implemented projectFacade

kubedo8:


war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
```diff
@@ -50,68 +66,404 @@ public void setCurrentProjectId(final String projectId) {
...
+    *
+    * @param organizationId
+    *       name of the organization
+    * @return map(id, name) values of all projects in an organization
+    */
+   public Map<String, String> readProjectsMap(final String organizationId) {
+      List<DataDocument> documents = dataStorage.searchIncludeAttrs(Project.COLLECTION_NAME, dataStorageDialect.fieldValueFilter(Project.ATTR_ORGANIZATION_ID, organizationId), Arrays.asList(Project.ATTR_PROJECT_NAME, Project.ATTR_PROJECT_ID));
+      Map<String, String> projectsMap = new HashMap<>();
+      documents.forEach(d -> projectsMap.put(d.getString(Project.ATTR_PROJECT_ID), d.getString(Project.ATTR_PROJECT_NAME)));
+      return projectsMap;
```
livthomas: What about using this instead of the previous three lines?

```java
return documents.stream().collect(Collectors.toMap(d -> d.getString(Project.ATTR_PROJECT_ID), d -> d.getString(Project.ATTR_PROJECT_NAME)));
```

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
```diff
@@ -50,68 +66,404 @@ public void setCurrentProjectId(final String projectId) {
...
+   /**
+    * Reads the project id according to its name
+    *
+    * @param projectName
+    *       name of the project
+    * @return id of the project
+    */
+   public String readProjectId(final String projectName) {
+      DataDocument document = dataStorage.readDocumentIncludeAttrs(Project.COLLECTION_NAME, projectNameFilter(projectName), Collections.singletonList(Project.ATTR_PROJECT_ID));
+      return document != null ? document.getString(Project.ATTR_PROJECT_ID) : null;
```
livthomas: Why not to use the other version of `getString` method which does the same thing?

```java
return document.getString(Project.ATTR_PROJECT_ID, null);
```

This can be used in many methods of `ProjectFacade` class.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
```diff
@@ -50,68 +66,404 @@ public void setCurrentProjectId(final String projectId) {
...
+    * Create new user role which consists of core roles
+    *
+    * @param projectId
+    *       Id of the project
+    * @param userRole
+    *       Unique name of user role
+    * @param coreRoles
+    *       List of the core roles to set
+    */
+   public void createRole(final String projectId, final String userRole, final List<String> coreRoles) {
```
livthomas: I think all role management methods should rather be in `UserFacade` or something similar and not in `ProjectFacade`.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
```diff
@@ -50,68 +66,404 @@ public void setCurrentProjectId(final String projectId) {
...
+         for (String ur : userRoles) {
+            if (roles.containsKey(ur) && roles.get(ur).contains(userRole)) {
+               return true;
+            }
+         }
+         return false;
+      }
+   }
 
+   private String userFilter(String projectId, String userName) {
```
livthomas: These filters should also be in some other class so they can be reused.

mongodb-storage/src/test/java/io/lumeer/mongodb/MongoDbStorageTest.java
```diff
@@ -521,38 +587,62 @@ public void testComplexArrayManipulation() throws Exception {
...
-      mongoDbStorage.removeItemFromArray(COLLECTION_COMPLEX_ARRAY_MANIPULATION, id, "n.a", d4);
-      fromDb = mongoDbStorage.readDocument(COLLECTION_COMPLEX_ARRAY_MANIPULATION, id);
+      mongoDbStorage.removeItemFromArray(COLLECTION_COMPLEX_ARRAY_MANIPULATION, documentIdFilter, "n.a", d4);
+      fromDb = mongoDbStorage.readDocument(COLLECTION_COMPLEX_ARRAY_MANIPULATION, documentIdFilter);
       assertThat(fromDb.getArrayList("n.a", DataDocument.class)).hasSize(2);
 
    }
 
    @Test
+   public void testttt() {
```
kubedo8: Ohh i forgot to delete this test.

api/src/main/java/io/lumeer/engine/api/LumeerConst.java
```diff
@@ -165,4 +171,28 @@ public static LinkDirection fromString(final String s) {
...
    }
+
+   public static class Project {
+
+      public static final String COLLECTION_NAME = "_system-project";
+      public static final String ATTR_PROJECT_ID = "project-id";
+      public static final String ATTR_ORGANIZATION_ID = "organization-id";
+      public static final String ATTR_PROJECT_NAME = "project-name";
+      public static final String ATTR_USERS = "users";
+      public static final String ATTR_USER = "user";
```
kubedo8: `Users` is name of the array field and `user` (annd also `user_roles` ) is the name of the field inside documents stored in this users array.

war/src/main/java/io/lumeer/engine/controller/ApplicationInitFacade.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class ApplicationInitFacade implements Serializable {
```
kubedo8: Ok

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
```diff
@@ -20,18 +20,25 @@
...
 import java.util.Map;
+import javax.annotation.PostConstruct;
 import javax.enterprise.context.RequestScoped;
 import javax.inject.Inject;
 
 /**
  * Manipulates with project related data.
  *
- * @author <a href="mailto:marvenec@gmail.com">Martin Večeřa</a>
+ * @author <a href="mailto:kubedo8@gmail.com">Jakub Rodák</a>
```
kubedo8: Ok

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
kubedo8: That looks nice, I wil change it!

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
kubedo8: Because it will throws `nullpointerexception` when document doesn't exists in Db.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
kubedo8: Yes i will do in next commit as we disccussed in meeting.

api/src/main/java/io/lumeer/engine/api/exception/ProjectNotFoundException.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------/
+ */
+package io.lumeer.engine.api.exception;
+
+/**
+ * @author <a href="mailto:kubedo8@gmail.com">Jakub Rodák</a>
+ */
+public class ProjectNotFoundException extends DbException {
```
kubedo8: This class is unused so i will completely delete it.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
kubedo8: It is only internal construct of dialect filter due to reusage in this class, so i don't agree. Are you sure?

api/src/main/java/io/lumeer/engine/api/data/DataStorage.java
```diff
@@ -136,87 +137,62 @@ default void connect(final StorageConnection connection, final String database,
...
-    * @param documentId
-    *       the id of the document
-    * @param version
-    *       the version of document
+    * @param filter
+    *       the filter to identify document
     * @return the DataDocument object representing the read document
     */
-   DataDocument readOldDocument(final String collectionName, final String documentId, final int version);
+   DataDocument readDocument(final String collectionName, final String filter);
```
kubedo8: I did this refactor due to lot of duplicated methods so I think that approach is better.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
livthomas: It depends on if we need to use these filters in some other classes. If not and every such query should go through other methods of `ProjectFacade`, they can stay here.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
livthomas: Sure. Sorry, I can see it now.

war/src/main/java/io/lumeer/engine/controller/DatabazeInitializer.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class DatabazeInitializer implements Serializable {
```
livthomas: `DatabaseInitializer`

api/src/main/java/io/lumeer/engine/api/LumeerConst.java
livthomas: This is very confusing. What about renaming `ATTR_USER` to `ATTR_USERNAME`?

kubedo8: I agree with that. So I will change it to some understandable format.

PR #35, opened 2017-03-29
Data filters instead of string conversion

marvec:
Please do not merge yet, just trying to run the tests...

### DIFF
--- a/mongodb-storage/src/main/java/io/lumeer/mongodb/MongoDbStorage.java
+++ b/mongodb-storage/src/main/java/io/lumeer/mongodb/MongoDbStorage.java
@@ -217,7 +217,7 @@ public class MongoDbStorage implements DataStorage {
 
    @Override
    public boolean collectionHasDocument(final String collectionName, final DataFilter filter) {
-      return database.getCollection(collectionName).find((Bson) filter.get()).limit(1).iterator().hasNext();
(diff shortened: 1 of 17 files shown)


marvec:
Please merge when the tests pass.

mongodb-storage/src/main/java/io/lumeer/mongodb/MongoDbStorage.java
```diff
@@ -215,8 +216,8 @@ public boolean hasCollection(final String collectionName) {
    }
 
    @Override
-   public boolean collectionHasDocument(final String collectionName, final String filter) {
-      return database.getCollection(collectionName).find(BsonDocument.parse(filter)).limit(1).iterator().hasNext();
+   public boolean collectionHasDocument(final String collectionName, final DataFilter filter) {
+      return database.getCollection(collectionName).find((Bson) filter.get()).limit(1).iterator().hasNext();
```
livthomas: The following code would probably be better in order to avoid casting:
```java
...find(filter.<Bson>get())...
```
But it is not very important and it can stay as it is.

marvec: That is actually a great idea! Will do that.

PR #36, opened 2017-03-30
Separate use roles managment from ProjectFacade

kubedo8:


war/src/test/java/io/lumeer/engine/controller/UserRoleFacadeIntegrationTest.java
```diff
@@ -0,0 +1,219 @@
...
+      final String coreRole1 = "coreRole1";
+      final String coreRole2 = "coreRole2";
+      final String coreRole3 = "coreRole3";
+
+      Map<String, List<String>> rolesProject1 = userRoleFacade.readRoles(organization, project1);
+      Map<String, List<String>> rolesProject2 = userRoleFacade.readRoles(organization, project2);
+      assertThat(rolesProject1).isEmpty();
+      assertThat(rolesProject2).isEmpty();
+
+      // create test
```
livthomas: You do not need to spend time editing this now but in the future try not to write such long test methods if they can be split into multiple smaller ones. It is then easier to find where the problem is when a test starts failing.

The tests can be split according to the comments you put there.

kubedo8: I put all things into one method, because they are connected. But it's good idea to separate it. I'll do it next time.

war/src/main/java/io/lumeer/engine/controller/UserRoleFacade.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.inject.Inject;
+
+/**
+ * Manipulates with user roles used in projects and organizations
+ */
+public class UserRoleFacade {
```
kubedo8: Ohhh I forgot to put scope to this Facade. I'll fix it!

war/src/main/java/io/lumeer/engine/controller/UserRoleFacade.java
```diff
@@ -0,0 +1,211 @@
...
+    * @param projectId
+    *       Id of the project
+    * @param userRole
+    *       Unique name of user role
+    * @param coreRoles
+    *       List of the core roles to set
+    */
+   public void createRole(final String organizationId, final String projectId, final String userRole, final List<String> coreRoles) {
+      DataDocument document = new DataDocument(LumeerConst.Project.UserRoles.ATTR_PROJECT_ID, projectId)
+            .append(LumeerConst.Project.UserRoles.ATTR_ORGANIZATION_ID, organizationId)
```
marvec: Project ID can be reused across organizations.

PR #39, opened 2017-04-01
OrganizationFacade #LMR-59 and org. roles management #LMR-58 implementation

matejperejda:


war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
```diff
@@ -59,6 +59,12 @@ public void init() {
                .append(LumeerConst.Project.UserRoles.ATTR_PROJECT_ID, LumeerConst.Index.ASCENDING)
                .append(LumeerConst.Project.UserRoles.ATTR_USER_ROLE, LumeerConst.Index.ASCENDING), true);
       }
+
+      if (!dataStorage.hasCollection(LumeerConst.Organization.COLLECTION_NAME)) {
+         dataStorage.createCollection(LumeerConst.Organization.COLLECTION_NAME);
+         dataStorage.createIndex(LumeerConst.Organization.COLLECTION_NAME, new DataDocument(LumeerConst.Organization.ATTR_ORG_ID, LumeerConst.Index.ASCENDING)
+               .append(dataStorageDialect.concatFields(LumeerConst.Organization.ATTR_USERS, LumeerConst.Organization.ATTR_USERS_USERNAME), LumeerConst.Index.ASCENDING), false);
+      }
```
kubedo8: You must create index on `Organization id` separately from `Users` and also create index on `Organization name`

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,463 @@
...
+    * @param organizationId
+    *       id of the new organization to create
+    * @param organizationName
+    *       name of the new organization to create
+    */
+   public void createOrganization(final String organizationId, final String organizationName) {
+      DataDocument document = new DataDocument(LumeerConst.Organization.ATTR_ORG_ID, organizationId)
+            .append(LumeerConst.Organization.ATTR_ORG_NAME, organizationName)
+            .append(LumeerConst.Organization.ATTR_ORG_DATA, new DataDocument())
+            .append(LumeerConst.Organization.ATTR_USERS, new DataDocument());
```
kubedo8: This is not neccessary, but you can keep it

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,463 @@
...
+    * @param organizationId
+    *       id of the organization
+    * @param infoAttribute
+    *       attribute of the organization additional info to drop
+    */
+   public void dropOrganizationInfoDataAttribute(final String organizationId, final String infoAttribute) {
+      DataDocument infoDocument = readOrganizationInfoData(organizationId);
+      infoDocument.remove(infoAttribute);
+      DataDocument document = new DataDocument(LumeerConst.Organization.ATTR_ORG_DATA, infoDocument);
+      dataStorage.updateDocument(LumeerConst.Organization.COLLECTION_NAME, document, organizationIdFilter(organizationId));
```
kubedo8: Are you sure it works? Update doesn't drop any attribute. You must must `dropAttribute` function.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,463 @@
...
+    * @param organizationId
+    *       id of the organization
+    * @param infoDataAttribute
+    *       attribute of the organization additional info
+    * @return value of the organization info attribute
+    */
+   public String readOrganizationInfoData(final String organizationId, final String infoDataAttribute) {
+      DataDocument document = dataStorage.readDocumentIncludeAttrs(LumeerConst.Organization.COLLECTION_NAME, organizationIdFilter(organizationId), Collections.singletonList(LumeerConst.Organization.ATTR_ORG_DATA));
+      DataDocument infoDataDocument = document.getDataDocument(LumeerConst.Organization.ATTR_ORG_DATA);
+      return infoDataDocument != null ? infoDataDocument.getString(infoDataAttribute) : null;
```
kubedo8: You should use previous function `readOrganizationInfoData` here, because it has duplicated implementation.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,463 @@
...
+
+   /**
+    * Drops the organization according to its id.
+    *
+    * @param organizationId
+    *       id of the given organization to drop
+    */
+   public void dropOrganization(final String organizationId) {
+      dataStorage.dropDocument(LumeerConst.Organization.COLLECTION_NAME, organizationIdFilter(organizationId));
+   }
```
kubedo8: We want here som checking if it can be achieved. But it needs other facades changes.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,463 @@
...
+   /**
+    * Resets additional info of the given organization.
+    *
+    * @param organizationId
+    *       id of the organization
+    */
+   public void resetOrganizationInfoData(final String organizationId) {
+      DataDocument defaultInfoDocument = new DataDocument(LumeerConst.Organization.ATTR_ORG_DATA, new DataDocument());
+      dataStorage.updateDocument(LumeerConst.Organization.COLLECTION_NAME, defaultInfoDocument, organizationIdFilter(organizationId));
+   }
```
kubedo8: I am not sure, if it works like this, did you test it?

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,463 @@
...
+      filter.put(LumeerConst.Organization.ATTR_ORG_ID, organizationId);
+      filter.put(dataStorageDialect.concatFields(LumeerConst.Organization.ATTR_USERS, LumeerConst.Organization.ATTR_USERS_USERNAME), userName);
+      return dataStorageDialect.multipleFieldsValueFilter(filter);
+   }
+
+   private DataFilter organizationFilterByUser(final String userName) {
+      Map<String, Object> filter = new HashMap<>();
+      filter.put(dataStorageDialect.concatFields(LumeerConst.Project.ATTR_USERS, LumeerConst.Project.ATTR_USERS_USERNAME), userName);
+      return dataStorageDialect.multipleFieldsValueFilter(filter);
+   }
```
kubedo8: Why do you use map here?

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,463 @@
...
+    *       id of the organization
+    * @param userName
+    *       name of the user to remove from organization
+    */
+   public void removeUserFromOrganization(final String organizationId, final String userName) {
+      List<String> userRoles = readUserRoles(organizationId, userName);
+      DataDocument userDocument = new DataDocument(LumeerConst.Organization.ATTR_USERS_USERNAME, userName)
+            .append(LumeerConst.Organization.ATTR_USERS_USER_ROLES, userRoles);
+      dataStorage.removeItemsFromArray(LumeerConst.Organization.COLLECTION_NAME, userFilter(organizationId, userName), LumeerConst.Organization.ATTR_USERS, Collections.singletonList(userDocument));
+   }
```
kubedo8: Unnessary reading of user roles. Look at `ProjectFacade.dropUserFromProject` method for way how to implement it correctly.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: Each method has its own test. Everything passes.

kubedo8: I realize that it's Ok...

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @kubedo8 Why is it not neccessary? It just creates a simple Organization entry with default attributes.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @kubedo8 Oh, I see. Thanks!

war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
matejperejda: @kubedo8 Thanks. Except of creating separated indexes for `id`, `users` and `name` I should let this code there?
```
dataStorage.createIndex(LumeerConst.Organization.COLLECTION_NAME, new DataDocument(LumeerConst.Organization.ATTR_ORG_ID, LumeerConst.Index.ASCENDING)
               .append(dataStorageDialect.concatFields(LumeerConst.Organization.ATTR_USERS, LumeerConst.Organization.ATTR_USERS_USERNAME), LumeerConst.Index.ASCENDING), false);
```

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @kubedo8 Ok, I'll put a `TODO` comment right there.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @kubedo8 Thanks, works perfect.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @kubedo8 You're right. It does not have to be there. I've already replaced that by `.fieldValueFilter()` method.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @kubedo8 Ok, I've done it.

war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
kubedo8: No because it doesn't guarantee unique `id` and `unique` name. And also when you searching on `name`, it should have his own index for fast response.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
kubedo8: Because when you put some values into this fields in future, it doesn't have effect whether this fields exit or not.

matejperejda: @kubedo8 Yes, that's corret. But you need to append at least the first `ATTR_ORG_NAME` attribute because of `organizationName`.

kubedo8: Of course. I mean only these empty attributes. `OrganizationId` and `OrganizationName` must be here.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,455 @@
...
+/**
+ * Manages organizations and keeps track of currently selected organization.
+ *
+ * @author <a href="mailto:marvenec@gmail.com">Martin Večeřa</a>
+ * @author <a href="mailto:mat.per.vt@gmail.com">Matej Perejda</a>
+ */
+@RequestScoped
+public class OrganizationFacade {
+
+   private static String organizationId = "ACME";
```
livthomas: This field cannot be static! Different users can use different organizations at the same time.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,455 @@
...
+
+   @Inject
+   @SystemDataStorage
+   private DataStorage dataStorage;
+
+   @Inject
+   private DataStorageDialect dataStorageDialect;
+
+   @Inject
+   private DatabaseInitializer databaseInitializer;
```
livthomas: I do not think you need to inject `DatabaseInitializer` here. It will get initialized anyway.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -0,0 +1,455 @@
...
+    * Renames a name of the given organization according to its id.
+    *
+    * @param organizationId
+    *       id of the given organization
+    * @param newOrganizationName
+    *       new name of the organization
+    */
+   public void renameOrganization(final String organizationId, final String newOrganizationName) {
+      DataDocument dataDocument = new DataDocument(LumeerConst.Organization.ATTR_ORG_NAME, newOrganizationName);
+      dataStorage.updateDocument(LumeerConst.Organization.COLLECTION_NAME, dataDocument, organizationIdFilter(organizationId));
```
livthomas: Should not this check if organization with such name does not exist yet? The same applies for `createOrganization` method. Or we can create some unique value constraint on that field. @kubedo8, what do you think?

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @livthomas Should the `"ACME" ` be a default value of `organizationId` ? Is it ok?

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
kubedo8: We have unique index on `organization name` so mongo will throw exception when this method try to update name to existing one. So here is question, should we catch this exception ? @livthomas

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
matejperejda: @livthomas I am not sure about that. @kubedo8 also uses this injection in the ProjectFacade implementation. https://github.com/Lumeer/engine/blob/devel/war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java#L58L59

kubedo8: I just forgot to delete it..

matejperejda: @kubedo8 Ok, I'm gonna delete it as well.

war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
```diff
@@ -69,6 +69,14 @@ public void init() {
                .append(LumeerConst.UserRoles.ATTR_PROJECT_ID, LumeerConst.Index.ASCENDING)
                .append(LumeerConst.UserRoles.ATTR_USER_ROLE, LumeerConst.Index.ASCENDING), true);
       }
+
+      if (!dataStorage.hasCollection(LumeerConst.Organization.COLLECTION_NAME)) {
+         dataStorage.createCollection(LumeerConst.Organization.COLLECTION_NAME);
+         dataStorage.createIndex(LumeerConst.Organization.COLLECTION_NAME, new DataDocument(LumeerConst.Organization.ATTR_ORG_ID, LumeerConst.Index.ASCENDING), true);
+         dataStorage.createIndex(LumeerConst.Organization.COLLECTION_NAME, new DataDocument(LumeerConst.Organization.ATTR_ORG_NAME, LumeerConst.Index.ASCENDING), true);
```
livthomas: According to the discussion on the meeting, the last parameter in this method should be `false` so there will not be unique index on organization names.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
livthomas: According to the discussion on the meeting, the organization name does not have to be unique so no changes are necessary here.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
livthomas: I think we should not provide any default value since the organization ID always needs to be set. And if it is not, we will spot the problem sooner.

PR #40, opened 2017-04-01
Add user check to ProjectFacade and refactor its test

kubedo8:
Add check for user uniqueness and refactor tests in case of potentional bug when methods are executed concurrently.
I forgot to append Issue number, so here it is: LMR121 (for merge).

### DIFF
--- a/war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
+++ b/war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
@@ -27,11 +27,16 @@ import io.lumeer.engine.api.data.DataStorageDialect;
 
 import java.io.Serializable;
 import javax.annotation.PostConstruct;
+import javax.ejb.Startup;
 import javax.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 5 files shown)


war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
```diff
@@ -40,6 +45,11 @@
    @Inject
    private DataStorageDialect dataStorageDialect;
 
+
+   public void init(@Observes @Initialized(RequestScoped.class) Object init) {
```
livthomas: Why is this method empty? Should not the initialization be done here instead of in `@PostConstruct` method? And what about changing `@Initialized` to use `ApplicationScoped` instead?

kubedo8: It can't be ApplicationScope, because it doesn't work this way (And I don't know why). This method ensure that `DatabaseInitializer` class is initialized so i put code into `PostConstruct` because I saw one solution programmed like this..

PR #41, opened 2017-04-03
introduced DataSort class

alicak:
Added DataSort class that supports sorting the way DataFilter supports filtering.

### DIFF
--- a/api/src/main/java/io/lumeer/engine/api/data/DataStorageDialect.java
+++ b/api/src/main/java/io/lumeer/engine/api/data/DataStorageDialect.java
@@ -33,9 +33,7 @@ public interface DataStorageDialect extends Serializable {
    DataDocument renameAttributeQuery(final String metadataCollection, final String collection, final String oldName, final String newName);
 
    DataDocument addRecentlyUsedDocumentQuery(final String metadataCollection, final String collection, final String id, final int listSize);
-
-   String sortStringForSearch(final String field, final int order); // produces string that can be used in dataStorage.search() method
-
+   
    // LinkingFacade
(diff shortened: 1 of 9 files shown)


api/src/main/java/io/lumeer/engine/api/data/DataStorageDialect.java
```diff
@@ -34,6 +34,8 @@
 
    DataDocument addRecentlyUsedDocumentQuery(final String metadataCollection, final String collection, final String id, final int listSize);
 
+   String sortStringForSearch(final String field, final int order); // produces string that can be used in dataStorage.search() method
+
    // LinkingFacade
```
kubedo8: This method is unused now, istn't it? (And probably unwanted)

alicak: Yes, you're right.

mongodb-storage/src/main/java/io/lumeer/mongodb/MongoDbStorageDialect.java
```diff
@@ -67,6 +69,10 @@ public DataDocument addRecentlyUsedDocumentQuery(final String metadataCollection
                               .append("$slice", listSize))));
    }
 
+   public String sortStringForSearch(final String field, final int order) {
+      return MongoUtils.convertBsonToJson(Sorts.descending(field));
```
livthomas: What is this method good for? I cannot see it used anywhere and it seems to ignore `order` parameter.

alicak: I'm sorry I didn't check it. It is forgotten method and I have already removed it.

PR #43, opened 2017-04-04
ViewFacade refactored

alicak:
Inspired by changes in CollectionMetadataFacade, I also changed ViewFacade. Checks of user rights are moved to ViewService, except filtering in `getAllViews` method - this will be edited when appropriate filter is available. I also divided views documents into collections by projectId - views from one project are in one collection.

### DIFF
--- a/war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
+++ b/war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
@@ -30,8 +30,11 @@ import io.lumeer.engine.util.ErrorMessageBuilder;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import javax.enterprise.context.SessionScoped;
 import javax.inject.Inject;
 
@@ -59,6 +62,9 @@ public class ViewFacade implements Serializable {
 
    @Inject
    private ProjectFacade projectFacade;
+
+   @Inject
+   private OrganisationFacade organisationFacade;
 
    /**
     * Creates initial metadata for the view
@@ -156,8 +162,7 @@ public class ViewFacade implements Serializable {
     * @return type of the given view
     */
    public String getViewType(int viewId) {
-      ViewDao metadata = getViewMetadata(viewId);
-      return metadata != null ? metadata.getType() : null;
+      return (String) getViewMetadataValue(viewId, LumeerConst.View.TYPE_KEY);
    }
 
    /**
@@ -177,8 +182,7 @@ public class ViewFacade implements Serializable {
     * @return view name
     */
    public String getViewName(int viewId) {
-      ViewDao metadata = getViewMetadata(viewId);
-      return metadata != null ? metadata.getName() : null;
+      return (String) getViewMetadataValue(viewId, LumeerConst.View.NAME_KEY);
    }
 
    /**
@@ -213,8 +217,7 @@ public class ViewFacade implements Serializable {
     * @return view description
     */
    public String getViewDescription(int viewId) {
-      ViewDao metadata = getViewMetadata(viewId);
-      return metadata != null ? metadata.getDescription() : null;
+      return (String) getViewMetadataValue(viewId, LumeerConst.View.DESCRIPTION_KEY);
    }
 
    /**
@@ -223,8 +226,7 @@ public class ViewFacade implements Serializable {
     * @return view configuration
     */
    public DataDocument getViewConfiguration(int viewId) {
-      ViewDao metadata = getViewMetadata(viewId);
-      return metadata != null ? metadata.getConfiguration() : null;
+      return (DataDocument) getViewMetadataValue(viewId, LumeerConst.View.CONFIGURATION_KEY);
    }
 
    /**
@@ -245,8 +247,7 @@ public class ViewFacade implements Serializable {
     * @return value of view configuration attribute
     */
    public Object getViewConfigurationAttribute(int viewId, String attributeName) {
-      ViewDao metadata = getViewMetadata(viewId);
-      return metadata != null ? metadata.getConfiguration().get(attributeName) : null;
+      return getViewConfiguration(viewId).get(attributeName);
    }
 
    /**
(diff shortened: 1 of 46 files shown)


war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -59,48 +57,62 @@
...
+         dataStorage.createCollection(metadataCollection());
          // we create indexes on frequently used fields
          int indexType = LumeerConst.Index.ASCENDING;
-         dataStorage.createIndex(LumeerConst.View.VIEW_METADATA_COLLECTION_NAME, new DataDocument(LumeerConst.View.VIEW_ID_KEY, indexType), false);
-         dataStorage.createIndex(LumeerConst.View.VIEW_METADATA_COLLECTION_NAME, new DataDocument(LumeerConst.View.VIEW_TYPE_KEY, indexType), false);
-         dataStorage.createIndex(LumeerConst.View.VIEW_METADATA_COLLECTION_NAME, new DataDocument(LumeerConst.View.VIEW_NAME_KEY, indexType), false);
+         dataStorage.createIndex(metadataCollection(), new DataDocument(LumeerConst.View.ID_KEY, indexType), false);
+         dataStorage.createIndex(metadataCollection(), new DataDocument(LumeerConst.View.TYPE_KEY, indexType), false);
+         dataStorage.createIndex(metadataCollection(), new DataDocument(LumeerConst.View.NAME_KEY, indexType), false);
       }
```
kubedo8: You should set this indexes as unique and remove `TYPE_KEY` index because of Cardinality.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -59,48 +57,62 @@
...
-            .append(LumeerConst.View.VIEW_ID_KEY, viewId)
-            .append(LumeerConst.View.VIEW_CREATE_USER_KEY, createUser)
-            .append(LumeerConst.View.VIEW_CREATE_DATE_KEY, new Date())
-            .append(LumeerConst.View.VIEW_TYPE_KEY, viewType) // sets view type to default
-            .append(LumeerConst.View.VIEW_CONFIGURATION_KEY, configuration != null ? configuration : new DataDocument());
+      DataDocument metadataDocument = new DataDocument(LumeerConst.View.NAME_KEY, originalViewName)
+            .append(LumeerConst.View.ID_KEY, viewId)
+            .append(LumeerConst.View.DESCRIPTION_KEY, description)
+            .append(LumeerConst.View.PROJECT_ID, projectFacade.getCurrentProjectId())
+            .append(LumeerConst.View.CREATE_USER_KEY, createUser)
```
kubedo8: You don't have to put `PROJECT_ID` into collection.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -113,111 +125,116 @@ public int createView(final String originalViewName, final String viewType, fina
...
+    *       view id
+    * @return object with view metadata
+    */
+   public ViewDao getViewMetadata(int viewId) {
+      DataDocument viewMetadata = getViewMetadataDocument(viewId);
+      if (viewMetadata == null) {
+         return null;
       }
+      return new ViewDao(viewMetadata);
    }
```
kubedo8: Inline this method like others.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -310,102 +286,103 @@ public Object getViewMetadataValue(int viewId, String metaKey) throws ViewMetada
...
+   }
+
+   /**
+    * @param projectId
+    *       project id
+    * @return name of view metadata collection for given project id
+    */
+   private String metadataCollection(String projectId) {
+      return LumeerConst.View.METADATA_COLLECTION_PREFIX + projectId;
+   }
```
kubedo8: You should use also organizationId to create collection. Or make separate collection for each organization and then use projectId as one of the field (don't forget to add projectId to index then)...

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -310,102 +286,103 @@ public Object getViewMetadataValue(int viewId, String metaKey) throws ViewMetada
...
-      if (viewList.isEmpty()) {
-         throw new ViewMetadataNotFoundException(ErrorMessageBuilder.viewMetadataNotFoundString(viewId));
-      }
+      return viewsDaos;
+   }
 
-      return viewList.get(0);
+   private boolean checkIfViewNameExists(String viewName) {
+      return dataStorage.getAttributeValues(metadataCollection(), LumeerConst.View.NAME_KEY).contains(viewName);
    }
```
kubedo8: Uneffective method. Use `collectionHasDocument` with filter on `NAME_KEY` field.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
alicak: I left it there since I create collection for organisation, not project.

war/src/main/java/io/lumeer/engine/rest/dao/ViewDao.java
```diff
@@ -28,6 +33,9 @@
  */
 public class ViewDao {
 
+   @Inject
+   private UserFacade userFacade;
```
livthomas: This must not be here! While this class is badly named, it was meant to be a [DTO](https://en.wikipedia.org/wiki/Data_transfer_object). It should only carry data  and if such information as currently logged user is required, it needs to be passed to it as a constructor argument when it is created.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -59,48 +60,65 @@
...
-      if (!dataStorage.hasCollection(LumeerConst.View.VIEW_METADATA_COLLECTION_NAME)) {
-         dataStorage.createCollection(LumeerConst.View.VIEW_METADATA_COLLECTION_NAME);
+   public int createView(
+         final String originalViewName,
+         final String viewType,
+         final String description,
+         final DataDocument configuration) throws ViewAlreadyExistsException {
+
+      if (!dataStorage.hasCollection(metadataCollection())) {
+         dataStorage.createCollection(metadataCollection());
```
livthomas: Do we want to check this every time this method is called? Isn't it an unnecessary database call? What about moving the initialization to `DatabaseInitializer`?

war/src/main/java/io/lumeer/engine/rest/dao/ViewDao.java
```diff
@@ -28,6 +33,9 @@
  */
 public class ViewDao {
```
livthomas: When doing refactoring, you can also rename this class to `ViewDto` or `ViewMetadata` which better describe what is it used for.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -247,165 +254,175 @@ public Object getViewConfigurationAttribute(int viewId, String attributeName) th
...
+    *       list of all views
+    * @return filtered list of views
+    */
+   private List<DataDocument> filterViewsForUser(List<DataDocument> views) {
+      List<DataDocument> filteredViews = new ArrayList<>();
+      String user = getCurrentUser();
+
+      for (DataDocument view : views) {
+         if (checkViewForRead(view.getInteger(LumeerConst.View.ID_KEY), user)) {
+            filteredViews.add(view);
```
livthomas: What about using Java Stream API?

```java
final String user = getCurrentUser();
return views.stream()
    .filter(view -> checkViewForRead(view.getInteger(LumeerConst.View.ID_KEY), user))
    .collect(Collectors.toList());
```

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -247,165 +254,175 @@ public Object getViewConfigurationAttribute(int viewId, String attributeName) th
...
+      return filteredViews;
    }
 
-   // gets info about one view without checking access rights
-   private DataDocument getViewMetadataWithoutAccessCheck(int viewId) throws ViewMetadataNotFoundException {
-      List<DataDocument> viewList = dataStorage.run(queryOneViewMetadata(viewId));
-      if (viewList.isEmpty()) {
-         throw new ViewMetadataNotFoundException(ErrorMessageBuilder.viewMetadataNotFoundString(viewId));
+   // creates list of daos from list of DataDocuments
+   private List<ViewDao> createListOfDaos(List<DataDocument> views) {
```
livthomas: DTOs...

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
alicak: I left it there as we have already discussed it.

PR #44, opened 2017-04-07
Enable restricted flag in configuration #LMR-130

livthomas:
You can find the configuration entry structure [on wiki](https://github.com/Lumeer/engine/wiki/Metadata-Structure#configurationfacade).

When merging this, do not squash the commits so it can easily be seen what changes are made in refactoring and which resolves LMR-130.

### DIFF
--- a/war/src/main/java/io/lumeer/engine/controller/configuration/ConfigurationManipulator.java
+++ b/war/src/main/java/io/lumeer/engine/controller/configuration/ConfigurationManipulator.java
@@ -27,10 +27,15 @@ import io.lumeer.engine.api.data.DataStorageDialect;
 import io.lumeer.engine.util.Resources;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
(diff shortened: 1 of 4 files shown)


war/src/main/java/io/lumeer/engine/controller/configuration/ConfigurationManipulator.java
```diff
@@ -150,8 +127,8 @@ public void resetConfiguration(final String collectionName, final String nameVal
     * @param value
     *       the Object value of the given key
     */
-   public void setConfiguration(final String collectionName, final String nameValue, final String key, final Object value) {
-      writeValueToDb(collectionName, nameValue, key, value);
+   public void setConfiguration(final String collectionName, final String nameValue, final String key, final Object value, final Map<String, Boolean> flags) {
```
marvec: Wouldn't be flags in form Set<String> be better? Allow null flags and an overloaded method variant without flags (assuming empty set).

war/src/test/java/io/lumeer/engine/controller/ConfigurationFacadeIntegrationTest.java
```diff
@@ -100,32 +100,32 @@ public void setUp() throws Exception {
...
+      final Optional<String> defaultValue = configurationFacade.getConfigurationString(DBHOST_KEY);
+      assertThat(defaultValue).contains(DEFAULT_DBHOST_VALUE);
 
       systemDataStorage.dropCollection(ConfigurationFacade.USER_CONFIG_COLLECTION);
       systemDataStorage.dropCollection(ConfigurationFacade.PROJECT_CONFIG_COLLECTION);
       systemDataStorage.dropCollection(ConfigurationFacade.ORGANISATION_CONFIG_COLLECTION);
 
       // #2 if the org. system collection has key-value
       fillSystemCollection(ConfigurationFacade.ConfigurationLevel.ORGANISATION);
-      assertThat(configurationFacade.getConfigurationString(DBHOST_KEY).get()).isEqualTo(DUMMY_DBHOST_VALUE);
```
livthomas: The assertion has been rewritten using optional assertion below.

PR #46, opened 2017-04-07
Added support for child attributes to DocumentFacade #LMR-97

kubedo8:


war/src/main/java/io/lumeer/engine/controller/DocumentFacade.java
```diff
@@ -324,10 +284,77 @@ public void dropAttribute(final String collectionName, final String documentId,
...
    public Set<String> getDocumentAttributes(final String collectionName, final String documentId) throws DbException {
       DataDocument dataDocument = dataStorage.readDocument(collectionName, dataStorageDialect.documentIdFilter(documentId));
-      Set<String> documentAttributes = new HashSet<>();
-      // filter out metadata attributes
-      documentAttributes.addAll(dataDocument.keySet().stream().filter(key -> !key.startsWith(LumeerConst.Document.METADATA_PREFIX)).collect(Collectors.toList()));
-      return documentAttributes;
+      return dataDocument != null ? getDocumentAttributes(dataDocument) : null;
+   }
+
+   private Set<String> getDocumentAttributes(DataDocument dataDocument) {
```
alicak: Maybe you could add some javadoc or just short comment also to private methods - for future reference.

livthomas: I do not think it is necessary to add JavaDoc for private methods. You usually want to break up your code to many private methods where every method does only a single thing and its purpose is obvious from its name. In that case the JavaDoc just makes the code less readable since you need to go through a lot of comments.

war/src/main/java/io/lumeer/engine/controller/DocumentFacade.java
```diff
@@ -93,26 +92,20 @@
...
-      // we add all document attributes to collection metadata
-      doc.keySet().stream().filter(attribute -> !LumeerConst.Document.METADATA_KEYS.contains(attribute)).forEach(attribute -> {
-         try {
-            collectionMetadataFacade.addOrIncrementAttribute(collectionName, attribute);
-         } catch (CollectionMetadataDocumentNotFoundException e) {
-            // nothing happens - if we don't find metadata, we just don't increment attribute
-         }
-      });
+
+      addOrIncrementAttributes(collectionName, documentCleaned);
```
livthomas: Why do we want to do this when creating a document?

PR #47, opened 2017-04-09
Added support for projects and organisations to CollectionFacade

alicak:
And I also changed exception handling - in case metadata is not found, exception is not thrown unless it is necessary.

### DIFF
--- a/war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
+++ b/war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
@@ -87,7 +87,7 @@ public class CollectionMetadataFacade implements Serializable {
 
    private ConstraintManager constraintManager;
 
-   private static final String METADATA_COLLECTION = LumeerConst.Collection.METADATA_COLLECTION;
+   private static final String METADATA_COLLECTION_PREFIX = LumeerConst.Collection.METADATA_COLLECTION_PREFIX;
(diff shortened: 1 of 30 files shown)


war/src/main/java/io/lumeer/engine/rest/dao/CollectionMetadata.java
```diff
@@ -42,10 +43,16 @@
    private String creator;
    private Date createDate;
 
+   // we put empty values here, so null pointer ex is not thrown when calling methods on the list or DataDocument
+   public CollectionMetadata() {
+      recentlyUsedDocumentIds = new ArrayList<>();
+      customMetadata = new DataDocument();
```
livthomas: These can be initialized above the same way as `attributes` map is.

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -921,8 +1047,19 @@ private String getCurrentUser() {
...
+   public String metadataCollection() {
+      return metadataCollection(organisationFacade.getOrganisationId());
+   }
+
+   /**
+    * @param orgId
+    *       organisation id
+    * @return name of metadata collection for given project id
+    */
+   public String metadataCollection(String orgId) {
```
livthomas: This method can be `static`.

Issue #2, opened 2016-11-17
Add audit logging or history tracking to configuration facade


marvec:
All configuration changes should be tracked and revertable.

Issue #1, opened 2016-10-29
Implement WebSocket security


marvec:
https://devcenter.heroku.com/articles/websocket-security

PR #48, opened 2017-04-13
Created ViewServiceIntegrationTest #LMR-124

matejperejda:


matejperejda:
@livthomas We've always done it the way. To avoid collections to be created and filled many times over and over again. So I've tested more methods or possible cases at once.

matejperejda:
@livthomas But there's no problem to change it. ;)

marvec:
A single test method should always test one thing. So actually putting there #1, #2... does not make sense. There are other ways how to create collections only once like with @BeforeClass. However, creating a collection is not time consuming operation. Please try to fix that in the future. For now, I am going to merge it...

PR #49, opened 2017-04-14
refactor ProjectFacade and OrganizationFacade #150

kubedo8:


war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
```diff
@@ -52,10 +52,6 @@ public void init() {
       if (!dataStorage.hasCollection(LumeerConst.Project.COLLECTION_NAME)) {
          dataStorage.createCollection(LumeerConst.Project.COLLECTION_NAME);
          dataStorage.createIndex(LumeerConst.Project.COLLECTION_NAME, new DataDocument(LumeerConst.Project.ATTR_ORGANIZATION_ID, LumeerConst.Index.ASCENDING)
-               .append(LumeerConst.Project.ATTR_PROJECT_NAME, LumeerConst.Index.ASCENDING), true);
-         dataStorage.createIndex(LumeerConst.Project.COLLECTION_NAME, new DataDocument(LumeerConst.Project.ATTR_ORGANIZATION_ID, LumeerConst.Index.ASCENDING)
-               .append(LumeerConst.Project.ATTR_PROJECT_ID, LumeerConst.Index.ASCENDING), true);
```
livthomas: Why is the index on project ID being removed?

war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
```diff
@@ -69,8 +65,6 @@ public void init() {
 
       if (!dataStorage.hasCollection(LumeerConst.Organization.COLLECTION_NAME)) {
          dataStorage.createCollection(LumeerConst.Organization.COLLECTION_NAME);
-         dataStorage.createIndex(LumeerConst.Organization.COLLECTION_NAME, new DataDocument(LumeerConst.Organization.ATTR_ORG_ID, LumeerConst.Index.ASCENDING), true);
```
livthomas: The same here.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -157,16 +148,38 @@ public void dropOrganizationMetadata(final String organizationId, final String m
...
+    *       new id for organization
+    * @throws OrganizationAlreadyExistsException
+    *       when organization with newOrganizationId already exists
+    */
+   public void updateOrganizationId(final String oldOrganizationId, final String newOrganizationId) throws OrganizationAlreadyExistsException {
+      if (dataStorage.collectionHasDocument(LumeerConst.Organization.COLLECTION_NAME, organizationIdFilter(newOrganizationId))) {
+         throw new OrganizationAlreadyExistsException(ErrorMessageBuilder.organizationAlreadyExistsString(organizationId));
+      }
+      DataDocument document = new DataDocument(LumeerConst.Organization.COLLECTION_NAME, newOrganizationId);
+      dataStorage.updateDocument(LumeerConst.Organization.COLLECTION_NAME, document, organizationIdFilter(oldOrganizationId));
```
livthomas: Isn't it somehow possible to detect if the organization with the same ID exists during `updateDocument` operation? I mean checking some specific MongoDB exception so we do not have to make two database calls.

kubedo8: Yes it is possible and it was done this way, but we have two options how to implement this. First is to have unique index on organization ID (same with project) and there's no need to check existency explicitly, but it cause duplicated index. Why? Because we have also compound index on organization ID + users so searching on organization ID can be covered by this index, but there's no mongo uniqueness checking and throwing exception. Second option is to remove index on organization ID and then we must explicitly check if it's already exists.
So what is better? Let's discuss it.. @livthomas

PR #51, opened 2017-04-15
Finished constraints #LMR-35

marvec:
It is not top notch, there are non-trivial code duplicates, but given the limited time I planned to spent here, I believe we are good to go.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -97,9 +97,32 @@
          <artifactId>org.jacoco.core</artifactId>
          <scope>test</scope>
       </dependency>
+
+      <!-- Testing -->
+      <dependency>
+         <groupId>junit</groupId>
+         <artifactId>junit</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>org.assertj</groupId>
+         <artifactId>assertj-core</artifactId>
+      </dependency>
+
    </dependencies>
    <build>
       <plugins>
+         <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+               <forkCount>1C</forkCount>
+               <reuseForks>true</reuseForks>
+               <redirectTestOutputToFile>false</redirectTestOutputToFile>
+               <systemPropertyVariables>
+                  <log4j.skipJansi>true</log4j.skipJansi>
+               </systemPropertyVariables>
+            </configuration>
+         </plugin>
          <plugin>
             <groupId>org.codehaus.mojo</groupId>
             <artifactId>license-maven-plugin</artifactId>
--- a/api/src/main/java/io/lumeer/engine/api/constraint/CaseConstraintType.java
+++ b/api/src/main/java/io/lumeer/engine/api/constraint/CaseConstraintType.java
@@ -19,12 +19,17 @@
  */
 package io.lumeer.engine.api.constraint;
 
+import java.text.DateFormat;
+import java.text.ParseException;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.function.BiFunction;
+import java.util.function.Function;
 
 /**
  * Makes sure the value is of the given case.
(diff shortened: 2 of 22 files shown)


api/src/main/java/io/lumeer/engine/api/constraint/ListConstraintType.java
```diff
@@ -38,9 +43,54 @@
...
    }
 
+   private BiFunction<Object, Class, Object> getEncodeFunction(final boolean oneOf) {
+      return (o, t) -> {
+         if (oneOf) {
+            if (t == null || t == String.class) {
+               return o.toString();
+            }
+         } else {
+            if (t == null || String[].class.isAssignableFrom(t)) {
```
livthomas: Is there always going to be an array of strings?

api/src/main/java/io/lumeer/engine/api/constraint/NumberConstraintType.java
```diff
@@ -51,88 +54,130 @@
...
    private NumberFormat numberFormat = NumberFormat.getNumberInstance();
-   private NumberFormat integerNumberFormat = NumberFormat.getNumberInstance();
+   private NumberFormat bigNumberFormat = NumberFormat.getNumberInstance();
+   private NumberFormat integerNumberFormat = NumberFormat.getIntegerInstance();
+   private NumberFormat currencyNumberFormat = NumberFormat.getNumberInstance();
 
    public NumberConstraintType() {
-      integerNumberFormat.setParseIntegerOnly(true);
+      ((DecimalFormat) currencyNumberFormat).setParseBigDecimal(true);
+      ((DecimalFormat) bigNumberFormat).setParseBigDecimal(true);
```
livthomas: What is the difference between `bigNumberFormat` and `currencyNumberFormat`?

marvec: For now there is none, in the future there might be. I am not sure if we want to add currency or so. So I kept that separated.

api/src/main/java/io/lumeer/engine/api/constraint/ListConstraintType.java
marvec: For tags type it must be.

api/src/main/java/io/lumeer/engine/api/constraint/CaseConstraintType.java
```diff
@@ -47,6 +52,30 @@
       return Collections.singleton(CONSTRAINT_PREFIX);
    }
 
+   static BiFunction<Object, Class, Object> getEncodeFunction() {
+      return (o, t) -> {
+         if (t != null && t != String.class) {
+            return null;
```
marvec: We accept only String.class request. So either the type is not specified or String is requested.

api/pom.xml
```diff
@@ -97,10 +97,33 @@
...
          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+               <forkCount>1C</forkCount>
+               <reuseForks>true</reuseForks>
+               <redirectTestOutputToFile>false</redirectTestOutputToFile>
+               <systemPropertyVariables>
+                  <log4j.skipJansi>true</log4j.skipJansi>
+               </systemPropertyVariables>
```
marvec: I just copy-pasted that from war module.

api/src/main/java/io/lumeer/engine/api/constraint/CaseConstraintType.java
livthomas: Oh, I see it now. It was just confusing at first sight.

api/src/main/java/io/lumeer/engine/api/constraint/DateTimeConstraintType.java
```diff
@@ -48,6 +51,38 @@
       return REGISTERED;
    }
 
+   private BiFunction<Object, Class, Object> getEncodeFunction(final DateFormat format) {
```
marvec: I did think about it. I even considered putting it to a separate class for all constraints. But something prevented me to do so. Will have a look...

api/pom.xml
marvec: I think that plugin configuration is not inherited from the parent pom. Moreover, I changed the values to run tests in multiple threads at the same time.

PR #53, opened 2017-04-27
OrganizationService

alicak:
Matej actually implemented almost everything. I changed REST URL paths, did some minor changes and corrected errors, and also added tests and javadoc. 
What needs to be done is to check rights and roles when doing operations, but to do that, we firstly need to redesign SecurityFacade.

### DIFF
--- a/war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
+++ b/war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
@@ -160,13 +160,9 @@ public class OrganizationFacade {
     *       id of the new organization to create
     * @param organizationName
     *       name of the new organization to create
+    * @throws OrganizationAlreadyExistsException when organization with given name already exists
     */
    public void createOrganization(final String organizationId, final String organizationName) throws OrganizationAlreadyExistsException {
-      // check if id is unique
-      if (readOrganizationName(organizationId) != null) {
-         throw new OrganizationAlreadyExistsException(ErrorMessageBuilder.organizationIdAlreadyExists(organizationId));
-      }
-
       // check if name is unique
       if (readOrganizationId(organizationName) != null) {
          throw new OrganizationAlreadyExistsException(ErrorMessageBuilder.organizationNameAlreadyExists(organizationName));
@@ -185,13 +181,9 @@ public class OrganizationFacade {
     *       id of the organization to change
     * @param newOrganizationId
     *       new id for organization
+    * @throws OrganizationAlreadyExistsException
     */
-   public void updateOrganizationId(final String oldOrganizationId, final String newOrganizationId) throws OrganizationAlreadyExistsException {
-      // check if id is unique
-      if (readOrganizationName(newOrganizationId) != null) {
-         throw new OrganizationAlreadyExistsException(ErrorMessageBuilder.organizationIdAlreadyExists(organizationId));
-      }
-
+   public void updateOrganizationId(final String oldOrganizationId, final String newOrganizationId) {
       DataDocument document = new DataDocument(LumeerConst.Organization.ATTR_ORG_ID, newOrganizationId);
       dataStorage.updateDocument(LumeerConst.Organization.COLLECTION_NAME, document, organizationIdFilter(oldOrganizationId));
    }
@@ -203,6 +195,7 @@ public class OrganizationFacade {
     *       id of the given organization
     * @param newOrganizationName
     *       new name of the organization
+    * @throws OrganizationAlreadyExistsException when organization with given name already exists
     */
    public void renameOrganization(final String organizationId, final String newOrganizationName) throws OrganizationAlreadyExistsException {
       // check if new name is unique
--- a/war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
+++ b/war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
@@ -25,7 +25,6 @@ import io.lumeer.engine.api.data.DataDocument;
 import io.lumeer.engine.api.data.DataFilter;
 import io.lumeer.engine.api.data.DataStorage;
 import io.lumeer.engine.api.data.DataStorageDialect;
-import io.lumeer.engine.api.exception.OrganizationAlreadyExistsException;
 import io.lumeer.engine.api.exception.UserAlreadyExistsException;
 import io.lumeer.engine.util.ErrorMessageBuilder;
 
@@ -90,18 +89,6 @@ public class OrganizationFacade {
    }
 
    /**
-    * Reads the organization id according to its name.
-    *
-    * @param organizationName
-    *       name of the organization
-    * @return id of the given organization
-    */
-   public String readOrganizationId(final String organizationName) {
-      DataDocument document = dataStorage.readDocumentIncludeAttrs(LumeerConst.Organization.COLLECTION_NAME, organizationNameFilter(organizationName), Collections.singletonList(LumeerConst.Organization.ATTR_ORG_ID));
-      return document != null ? document.getString(LumeerConst.Organization.ATTR_ORG_ID) : null;
-   }
-
-   /**
     * Reads the specific organization metadata.
     *
     * @param organizationId
@@ -160,14 +147,8 @@ public class OrganizationFacade {
     *       id of the new organization to create
     * @param organizationName
     *       name of the new organization to create
-    * @throws OrganizationAlreadyExistsException when organization with given name already exists
     */
-   public void createOrganization(final String organizationId, final String organizationName) throws OrganizationAlreadyExistsException {
-      // check if name is unique
-      if (readOrganizationId(organizationName) != null) {
-         throw new OrganizationAlreadyExistsException(ErrorMessageBuilder.organizationNameAlreadyExists(organizationName));
-      }
-
+   public void createOrganization(final String organizationId, final String organizationName) {
       DataDocument document = new DataDocument(LumeerConst.Organization.ATTR_ORG_ID, organizationId)
             .append(LumeerConst.Organization.ATTR_ORG_NAME, organizationName)
             .append(LumeerConst.Organization.ATTR_ORG_DATA, new DataDocument());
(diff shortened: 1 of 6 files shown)


war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,389 @@
...
+@Path("/organizations")
+@RequestScoped
+public class OrganizationService implements Serializable {
+
+   private static final long serialVersionUID = 3125094059637285633L;
+
+   @Inject
+   private OrganizationFacade organizationFacade;
+
+   // TODO: RequestScoped?
```
livthomas: There is no need to create a new instance of this class with every request so it can be `ApplicationScoped`.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,389 @@
...
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * @author <a href="mailto:mat.per.vt@gmail.com">Matej Perejda</a>
+ * @author <a href="alica.kacengova@gmail.com">Alica Kačengová</a>
```
livthomas: We should rather describe what is this class good for and avoid `@author` tags. The explanation can be found [here](http://vojtechruzicka.com/stop-using-javadoc-author-tag/).

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,389 @@
...
+   }
+
+   /**
+    * @param organizationName organization name
+    * @return id of given organization
+    */
+   @GET
+   @Path("/name/{organizationName}")
+   @Produces(MediaType.APPLICATION_JSON)
+   public String getOrganizationId(final @PathParam("organizationName") String organizationName) {
```
livthomas: This is not needed as discussed above.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,389 @@
...
+      }
+      return organizationFacade.readOrganizationId(organizationName);
+   }
+
+   /**
+    * @param organizationId organization id
+    * @return name of given organization
+    */
+   @GET
+   @Path("/id/{organizationId}")
```
livthomas: Since organization name method is not needed, we can skip `id` part here. This applies to other methods as well.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,389 @@
...
+   }
+
+   /**
+    * @param organizationId organization id
+    * @param userName user name
+    * @param userRoles list of user roles
+    * @throws UserAlreadyExistsException when user with given name already exists
+    */
+   @POST
+   @Path("/id/{organizationId}/users/{userName}") // TODO: aku adresu?
```
livthomas: The address is good.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,389 @@
...
+      }
+      organizationFacade.addRolesToUser(organizationId, userName, userRoles);
+   }
+
+   /**
+    * @param organizationId organization id
+    * @param userName user name
+    * @param userRoles list of roles
+    */
+   @PUT
```
livthomas: Why don't we use `@DELETE` here?

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -89,6 +90,18 @@ public String readOrganizationName(final String organizationId) {
    }
 
    /**
+    * Reads the organization id according to its name.
+    *
+    * @param organizationName
+    *       name of the organization
+    * @return id of the given organization
+    */
+   public String readOrganizationId(final String organizationName) {
```
alicak: Task description https://lumeer.myjetbrains.com/youtrack/issue/LMR-135 says "read organization id according to organization name" so I thought it has to be unique.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -325,7 +351,11 @@ public void addUserToOrganization(final String organizationId, final String user
     *       When user already exists in project
     */
    public void addUserToOrganization(final String organizationId, final String userName) throws UserAlreadyExistsException {
-      addUserToOrganization(organizationId, userName, readDefaultRoles(organizationId));
+      List<String> roles = readDefaultRoles(organizationId);
+      if (roles == null) {
+         roles = Collections.emptyList();
```
alicak: When there are no default roles set, that attribute value does not exist in organization metadata. And if we then add new user, this null is read as default roles and is added to that new user. And then an exception is thrown when reading user and its roles because empty list is expected, not null.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
alicak: `@author` tags are generated automatically when creating new class. I just added my name.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
livthomas: The description is outdated. It has been changed in [this commit](https://github.com/Lumeer/engine/commit/c16febe143b360da6967004c0b9461fd8727bf32).

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
alicak: Because when I was writing test, I realized that  `@DELETE` cannot carry anything in message body, as here https://github.com/alicak/engine/blob/dea72f86d1052a5138025e1ebc9705639d061b8f/war/src/test/java/io/lumeer/engine/rest/OrganizationServiceIntegrationTest.java#L166. Also, `@DELETE` should be used when deleting the resource, but we are just partially updating it, so I used `@PUT`.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
livthomas: Hmm, we should probably put an empty list there when creating an organization. But also return an empty list in `readDefaultRoles` method instead of `null`.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
livthomas: But having POST for adding roles and PUT for removing them is very confusing. We should rather have a single PUT method which will be idempotent and will override all roles. But the underlying facade methods might need to be changed as well to support this.

The other solution might be adding and removing only a single roles. In that case it can be specified in the path.

But we do not know what exactly will be needed in the UI. Maybe both solutions can be implemented if possible. What do you think?

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,371 @@
...
+      }
+      return organizationFacade.readOrganizationName(organizationId);
+   }
+
+   /**
+    * @param organizationId organization id
+    * @param organizationName organization name
+    */
+   @POST
+   @Path("/{organizationId}/name/{organizationName}")
```
livthomas: This should be changed to `/{organizationId}` since you create an organization not just its name. The name should be passed in the body.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,371 @@
...
+         throw new IllegalArgumentException();
+      }
+      organizationFacade.dropOrganizationInfoDataAttribute(organizationId, attributeName);
+   }
+
+   /**
+    * Drops all additional info.
+    * @param organizationId organization id
+    */
+   @PUT
```
livthomas: This should be DELETE as well.

war/src/main/java/io/lumeer/engine/rest/OrganizationService.java
```diff
@@ -0,0 +1,371 @@
...
+         throw new IllegalArgumentException();
+      }
+      return organizationFacade.readDefaultRoles(organizationId);
+   }
+
+   /**
+    * @param organizationId organization id
+    * @param userRoles list of default roles
+    */
+   @POST
```
livthomas: Does this method override all the default roles? If so, it should rather be PUT in my opinion.

war/src/main/java/io/lumeer/engine/util/ErrorMessageBuilder.java
```diff
@@ -63,6 +63,9 @@ private ErrorMessageBuilder() {
    private static final String ERROR_USER_ALREADY_EXISTS_IN_PROJECT = "User \"{0}\" already exists in project \"{1}\"";
    private static final String ERROR_USER_ALREADY_EXISTS_IN_ORGANIZATION = "User \"{0}\" already exists in organization \"{1}\"";
 
+   private static final String ERROR_ORGANIZATION_NAME_ALREADY_EXISTS = "Organization with name \"{0}\" already exists.";
```
livthomas: Is this still needed?

war/src/main/java/io/lumeer/engine/util/ErrorMessageBuilder.java
```diff
@@ -175,4 +178,15 @@ public static String userAlreadyExistsInOrganizationString(String userName, Stri
       return MessageFormat.format(ERROR_USER_ALREADY_EXISTS_IN_ORGANIZATION, userName, organizationId);
    }
 
+   public static String organizationNameAlreadyExists(String name) {
```
livthomas: This is no longer needed I guess.

PR #56, opened 2017-05-02
SecurityFacade refactor

alicak:
It's not complete nor tested nor used in other facades, but I wanted to check if design of methods is ok. I was thinking about how much should they be parametrized but I decided that it would be better to use one method for every role.


war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
```diff
@@ -51,8 +52,285 @@
...
+   public boolean hasViewRole(String viewId, String projectId, String user, String role) {
+      DataDocument doc = dataStorage.readDocumentIncludeAttrs(
+            LumeerConst.Security.ROLES_COLLECTION_NAME,
+            dataStorageDialect.multipleFieldsValueFilter(viewFilter(viewId, projectId)),
+            Arrays.asList(dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY, role)));
+
+      return checkRole(doc, user, role);
+   }
+
+   private boolean checkRole(DataDocument rolesDocument, String user, String role) {
```
livthomas: We should not pass `user` as a parameter here but get it from `UserFacade.getUserEmail()` instead.

livthomas:
Methods like `initialize*Roles` should be moved to `DatabaseInitializer` and called in facades where organizations/projects/collections/views are created.

war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
```diff
@@ -51,9 +52,377 @@
...
+   public boolean hasProjectRole(String projectId, String role) {
+      DataDocument doc = dataStorage.readDocumentIncludeAttrs(
+            LumeerConst.Security.ROLES_COLLECTION_NAME,
+            dataStorageDialect.multipleFieldsValueFilter(projectFilter(projectId)),
+            Arrays.asList(dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY, role)));
+
+      return checkRole(doc, userFacade.getUserEmail(), role);
+   }
+
+   public boolean hasCollectionRole(String collectionName, String projectId, String role) {
```
livthomas: Just a detail but I would rather but `projectId` parameter first since it is higher in the hierarchy.

war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
```diff
@@ -51,9 +52,377 @@
...
+      DataDocument doc = dataStorage.readDocumentIncludeAttrs(
+            LumeerConst.Security.ROLES_COLLECTION_NAME,
+            dataStorageDialect.multipleFieldsValueFilter(collectionFilter(collectionName, projectId)),
+            Arrays.asList(dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY, role)));
+
+      return checkRole(doc, userFacade.getUserEmail(), role);
+   }
+
+   public boolean hasViewRole(int viewId, String projectId, String role) {
```
livthomas: The same here.

war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
```diff
@@ -51,9 +52,377 @@
...
+            Arrays.asList(dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY, role)));
+
+      return checkRole(doc, userFacade.getUserEmail(), role);
+   }
+
+   private boolean checkRole(DataDocument rolesDocument, String user, String role) {
+      List<String> groupsForUser = Collections.emptyList(); // TODO: user UsersGroupFacade
+      String key = dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY, role);
+
+      // does not work :-(
```
livthomas: What are we going to do about it? :)

livthomas:
And what about creating indexes on these collections? There should be methods for it added to `DatabaseInitializer` and called when a project/organization is created.

war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
alicak: I don't know if it is necessary to do something right now, as the workaround works, but I told Kubo about it.

war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
```diff
@@ -51,9 +52,296 @@
...
+
+   private String userOrGroupKey(String user) {
+      return user != null ? LumeerConst.Security.USERS_KEY : LumeerConst.Security.GROUP_KEY;
+   }
+
+   private String userOrGroupName(String user, String group) {
+      return user != null ? user : group;
+   }
+
+   /********* OLD **********/
```
livthomas: Do not we want to remove the old stuff? We can always find it in the Git history if needed.

alicak: I haven't removed it yet because then I would have to change many other facades which use those methods. I will of course do it once all new methods are finished.

war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
```diff
@@ -310,22 +316,22 @@ private void removeViewRole(String projectId, int viewId, String user, String gr
 
    private Map<String, Object> projectFilter(String projectId) {
```
livthomas: `projectId` parameters in such methods should be renamed to `projectCode` not to confuse anybody in the future.

PR #57, opened 2017-05-07
Created UserGroupFacade #LMR-171

kubedo8:


mongodb-storage/src/main/java/io/lumeer/mongodb/MongoDbStorageDialect.java
```diff
@@ -65,11 +66,18 @@ public DataDocument addRecentlyUsedDocumentQuery(final String metadataCollection
             .append("update", new DataDocument()
                   .append("$push", new DataDocument()
                         .append(LumeerConst.Collection.RECENTLY_USED_DOCUMENTS_KEY, new DataDocument()
-                              .append("$each", Arrays.asList(id))
+                              .append("$each", Collections.singletonList(id))
                               .append("$position", 0)
                               .append("$slice", listSize))));
    }
 
+   public DataDocument[] usersOfGroupAggregate(final String organization, final String group) {
```
livthomas: This should be made more generic. Getting nested array will be needed in the other parts of the system as well. Moreover, we should not put domain-specific methods to the database storage implementation.

kubedo8: I'm afraid that it's hard to make generic method for situations like this. Aggregate phases (and als their count) can be different in each situation. Of course there is a way how to do it more generic, but the method will take lot of params and inside the method we will need loot of conditions...

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -186,8 +190,12 @@ public void renameOrganization(final String organizationId, final String newOrga
     *       id of the given organization to drop
     */
    public void dropOrganization(final String organizationId) {
-      dataStorage.dropDocument(LumeerConst.Organization.COLLECTION_NAME, organizationIdFilter(organizationId));
       // TODO: check cannot be implemented without other facades changes
+      DataDocument document = dataStorage.readDocumentIncludeAttrs(LumeerConst.Organization.COLLECTION_NAME, organizationIdFilter(organizationId), Collections.singletonList(LumeerConst.Document.ID));
```
livthomas: There is a new method for this introduced in commit 28aec05.

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
```diff
@@ -0,0 +1,298 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+
+/**
+ * Stores users and their groups
+ */
+@RequestScoped
```
livthomas: Why `@RequestScoped`? There is no need for this. It should be `@SessionScoped`. In the future, it can even be `@ApplicationScoped` if we find a way to make system data storage application-scoped bean.

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
```diff
@@ -0,0 +1,298 @@
...
+    *
+    * @param organization
+    *       Id of the organization.
+    * @param user
+    *       User identificator.
+    * @param group
+    *       The name of group.
+    */
+   public void addGroupToUser(final String organization, final String user, final String group) {
+      addGroupsToUser(organization, user, Collections.singletonList(user));
```
livthomas: Users are added to groups not vice versa. So this method should be called `addUserToGroup`. Also there is a bug. You pass `user` as a group parameter.

However, I think this can be made even better by merging it with the `addGroupsToUser` to a single method:
```java
public void addUserToGroup(final String organization, final String user, final String... groups) {
   dataStorage.addItemsToArray(UserGroup.COLLECTION_NAME, userFilter(organization, user), userParam(), Arrays.asList(groups));
}
```
In this case you only need one method but can use it both ways.

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
```diff
@@ -0,0 +1,298 @@
...
+    * Remove group from user
+    *
+    * @param organization
+    *       Id of the organization.
+    * @param user
+    *       User identificator.
+    * @param group
+    *       The name of group.
+    */
+   public void removeGroupFromUser(final String organization, final String user, final String group) {
```
livthomas: Exactly the same comments as above: naming and single method

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
```diff
@@ -0,0 +1,298 @@
...
+
+   /**
+    * Remove groups from organization
+    *
+    * @param organization
+    *       Id of the organization.
+    * @param groups
+    *       The list of group names.
+    */
+   public void removeGroups(final String organization, final List<String> groups) {
```
livthomas: Again, we only need one method with `String... groups` parameter.

mongodb-storage/src/main/java/io/lumeer/mongodb/MongoDbStorage.java
```diff
@@ -382,26 +382,34 @@ public void dropAttribute(final String collectionName, final DataFilter filter,
...
       }
-      removeItemsFromArrayInternal(collectionName, filter, attributeName, items);
+      removeItemsFromArrayInternal(collectionName, filter, attributeName, items, multi);
    }
 
-   private <T> void removeItemsFromArrayInternal(final String collectionName, final DataFilter filter, final String attributeName, final List<T> items) {
-      database.getCollection(collectionName).updateOne(filter.<Bson>get(), pullAll(attributeName, items));
+   private <T> void removeItemsFromArrayInternal(final String collectionName, final DataFilter filter, final String attributeName, final List<T> items, final boolean multi) {
+      if (multi) {
+         database.getCollection(collectionName).updateMany(filter.<Bson>get(), pullAll(attributeName, items));
```
livthomas: Is this `multi` parameter really necessary? Why cannot we always use `updateMany` method?

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
```diff
@@ -0,0 +1,298 @@
...
+   /**
+    * Read users for specific group.
+    *
+    * @param organization
+    *       Id of the organization.
+    * @param group
+    *       Group name.
+    * @return The list of user names.
+    */
+   public List<String> getUsersOfGroup(final String organization, final String group) {
```
livthomas: I would call this `getUsersInGroup` so it won't be confused with `getGroupsOfUser` method.

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
```diff
@@ -0,0 +1,298 @@
...
+    * Read groups in organization.
+    *
+    * @param organization
+    *       Id of the organization.
+    * @return The list of group names.
+    */
+   public List<String> getGroups(final String organization) {
+      DataDocument document = dataStorage.readDocumentIncludeAttrs(Group.COLLECTION_NAME, organizationFilterGroups(organization), Collections.singletonList(Group.ATTR_GROUPS));
+      List<String> groups = document.getArrayList(Group.ATTR_GROUPS, String.class);
+      return groups != null ? groups : Collections.emptyList();
```
livthomas: Just a question I have probably asked before. Why don't we return an empty list directly in `getArrayList` so you never need to check if it is not `null`? It would save us some lines of code.

livthomas:
Also merge conflicts need to be resolved because of changes made by @alicak.

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
kubedo8: I realized that bug immediately after commit, but I didn't make another commit, because i knew, that there will be some other changes too..

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
kubedo8: Okay, that's good idea.

war/src/main/java/io/lumeer/engine/controller/UserGroupFacade.java
```diff
@@ -0,0 +1,261 @@
...
+
+   /**
+    * Adds one or more groups to organization
+    *
+    * @param organization
+    *       Id of the organization.
+    * @param groups
+    *       The one or more group names.
+    */
+   public void addGroup(final String organization, final String... groups) {
```
livthomas: I think all these methods where there is `String... groups` parameter should be called `addGroups` etc. (plural form).

war/src/main/java/io/lumeer/engine/controller/configuration/ConfigurationManipulator.java
```diff
@@ -235,7 +235,7 @@ private void writeValueToDb(final String collectionName, final String nameValue,
       configValues.put(key, value);
 
       List<String> flagsArray = configDocument.getArrayList(FLAGS_KEY, String.class);
-      if (flagsArray == null) {
+      if (flagsArray.isEmpty()) {
          flagsArray = new ArrayList<>();
```
livthomas: This if statement can be removed.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -198,8 +202,12 @@ public void renameOrganization(final String organizationId, final String newOrga
     *       id of the given organization to drop
     */
    public void dropOrganization(final String organizationId) {
-      dataStorage.dropDocument(LumeerConst.Organization.COLLECTION_NAME, organizationIdFilter(organizationId));
       // TODO: check cannot be implemented without other facades changes
```
livthomas: Can we remove this TODO comment? Or is there still something that needs to be done here?

PR #59, opened 2017-05-09
Add search query suggestions #LMR-164

livthomas:


war/src/main/java/io/lumeer/engine/controller/search/QuerySuggester.java
```diff
@@ -0,0 +1,186 @@
...
+
+   List<DataDocument> suggestAttributes(String text) {
+      String[] parts = text.split("\\.", 2);
+      if (parts.length < 2) {
+         return suggestCollectionsForAttribute(text);
+      }
+
+      String collectionName = parts[0];
+      String attributePart = parts[1];
+
```
kubedo8: What if we want to find attribute in each collection? Or is it necessary to set collection to query in GUI?

kubedo8:
We could do this search quite sophisticated: find all occurences, sort by most relevant and then return the best ones... but that's maybe out of scope now.

war/src/main/java/io/lumeer/engine/controller/search/QuerySuggester.java
livthomas: I thought it would be an overkill to search for all attributes of all collections but as I am thinking of it know there will probably not be so many of them so it might not be that inefficient.

api/src/main/java/io/lumeer/engine/api/data/DataStorage.java
```diff
@@ -356,6 +356,21 @@ default void connect(final StorageConnection connection, final String database,
...
+    *       the name of the collection where to search
+    * @param filter
+    *       the query predicate. If unspecified, then all documents in the collection will match the predicate.
+    * @param attributes
+    *       attributes included in the result (excluding _id)
+    * @param limit
+    *       the maximum number of documents to return (0 means no limit)
+    * @return the list of the found documents
+    */
+   List<DataDocument> search(String collectionName, DataFilter filter, List<String> attributes, int limit);
```
livthomas: Oh, I somehow overlooked that one. Nice catch! I will merge them.

war/src/main/java/io/lumeer/engine/controller/search/QuerySuggester.java
livthomas: I am going to merge this pull-request without resolving this particular concern since it would be difficult (and probably inefficient as well) to implement searching in all attributes of all collections unless we change the way attributes are stored in the database.

PR #60, opened 2017-05-15
Resolved project/organization id and code

alicak:


war/src/main/java/io/lumeer/engine/rest/BatchService.java
```diff
@@ -70,8 +70,8 @@
 
    @PostConstruct
    public void init() {
-      organizationFacade.setOrganizationId(organisationId);
-      projectFacade.setCurrentProjectId(projectId);
+      organizationFacade.setOrganizationCode(organisationId);
+      projectFacade.setCurrentProjectCode(projectId);
```
livthomas: `organizationId` and `projectId` attributes in this and other service classes need to be renamed to `organizationCode` and `projectCode` (or just `organization` and `project`).

livthomas:
One last suggestion for improvement that has just came to my mind. We should cache `projectId` in `ProjectFacade` and `organizationId` in `OrganizationFacade` because we are making too many database calls right now. The first call could store it in an attribute and all the subsequent calls will get it from the attribute and will not make any unnecessary database calls.

PR #61, opened 2017-05-19
UserGroupService

alicak:
There are problems with two tests, detailed description is provided in comments in tests code. The first has problem with json parsing and the second is really strange - even facade is behaving in unexpected way. Maybe Kubo can find it out. I am also not sure about the scope of the class.


war/src/test/java/io/lumeer/engine/rest/UserGroupServiceIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+   }
+
+   // TODO: reading map from response does not work (JsonMappingException) and I don't know why"
+   @Test
+   public void testGetUsersAndGroups() throws Exception {
+      final Client client = ClientBuilder.newBuilder().build();
+
+      Response response = client.target(TARGET_URI).path(PATH_PREFIX + "users").request(MediaType.APPLICATION_JSON).buildGet().invoke();
+      Map<String, List<String>> usersAndGroups = response.readEntity(Map.class); // does not work
+      assertThat(usersAndGroups).isEmpty();
```
kubedo8: Try `  Map<String, List<String>> usersAndGroups = response.readEntity(new GenericType<Map<String, List<String>>>() {});` ... It should work

war/src/test/java/io/lumeer/engine/rest/UserGroupServiceIntegrationTest.java
```diff
@@ -0,0 +1,251 @@
...
+      String group = "GetUsersInGroup";
+      String user1 = "user 1";
+      String user2 = "user 2";
+
+      userGroupFacade.addUser(organization, user1, group);
+      userGroupFacade.addUser(organization, user2, group);
+
+      // even command from facade does not work the way I expect, and I can't find out why... :-(
+      assertThat(userGroupFacade.getUsersInGroup(organization, group)).containsOnly(user1, user2);
+
```
kubedo8: You forgot one important thing -> Create organization. In this case there isn't Document in DB that match your criteria to put user into it. This should be done in each method (like in `UserGroupFacadeIntegrationTest`).

alicak: I drop and create organization before every test in `init()` method, so the problem is somewhere else.

war/src/test/java/io/lumeer/engine/rest/UserGroupServiceIntegrationTest.java
alicak: I've seen it in some test but it didn't work... I suppose it's because there is a collection in collection?

war/src/test/java/io/lumeer/engine/rest/UserGroupServiceIntegrationTest.java
kubedo8: Oh I see. So I think that problem is in your recent changes in `UserGroupFacade` (changing method params from id to organization code).

kubedo8: Oh I see. So this is so strange... try to repair `UserGroupFacadeIntegrationTest` to put organization code as parameter instead of id to check if there is everything okay.

war/src/main/java/io/lumeer/engine/rest/UserGroupService.java
```diff
@@ -0,0 +1,134 @@
...
+
+   @Inject
+   private OrganizationFacade organizationFacade;
+
+   @PathParam("organization")
+   private String organizationCode;
+
+   @POST
+   @Path("/users/{user}")
+   @Consumes(MediaType.APPLICATION_JSON)
```
livthomas: You can put all these `@Consumes` and `@Produces` annotations on the class level so they do not need to be specified on each method.

war/src/main/java/io/lumeer/engine/rest/UserGroupService.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * @author <a href="alica.kacengova@gmail.com">Alica Kačengová</a>
+ */
+
+@Path("/{organization}/groups")
```
livthomas: This is not a good path for both users and groups methods. It might be confusing. That is why I would suggest splitting this service into two different classes `UserService` and `GroupService`.

`UserService` would provide the following methods:
```
GET    /{organization}/users/
POST   /{organization}/users/{user}
DELETE /{organization}/users/{user}
GET    /{organization}/users/{user}/groups
```

And `GroupService` would provide these methods:
```
GET    /{organization}/groups
POST   /{organization}/groups
DELETE /{organization}/groups/{group}
GET    /{organization}/groups/{group}/users
PUT    /{organization}/groups/{group}/users/{user}
DELETE /{organization}/groups/{group}/users/{user}
```

war/src/main/java/io/lumeer/engine/rest/UserGroupService.java
```diff
@@ -0,0 +1,134 @@
...
+   }
+
+   @POST
+   @Path("/")
+   @Consumes(MediaType.APPLICATION_JSON)
+   public void addGroups(final List<String> groups) {
+      userGroupFacade.addGroups(organizationCode, toArray(groups));
+   }
+
+   @PUT // PUT because DELETE cannot take any arguments in message body
```
livthomas: What about using `@Delete` here and passing groups as comma-separated list of group names which then will be parsed and sent to facade as an array?
```java
public void removeGroups(@QueryParam("groups") String groups) {
   userGroupFacade.removeGroups(organizationCode, groups.split(","));
}
```
Or even better solution would be to allow only a single group creation/deletion. If we need to multiple with more groups at the same, we can still add this functionality in the future.

PR #62, opened 2017-05-22
Changed how org and project return basic data #LMR-157

kubedo8:


war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -80,13 +80,12 @@ public void setOrganizationCode(final String organizationCode) {
...
+    * Reads  all organizations in the system.
     *
-    * @return map of values (code, name) of all organizations in the system
+    * @return list of organizations
     */
-   public Map<String, String> readOrganizationsMap() {
-      List<DataDocument> documents = dataStorage.search(LumeerConst.Organization.COLLECTION_NAME, null, Arrays.asList(LumeerConst.Organization.ATTR_ORG_NAME, LumeerConst.Organization.ATTR_ORG_CODE));
-      return documents.stream().collect(Collectors.toMap(d -> d.getString(LumeerConst.Organization.ATTR_ORG_CODE), d -> d.getString(LumeerConst.Organization.ATTR_ORG_NAME)));
+   public List<DataDocument> readOrganizations() {
+      return dataStorage.search(LumeerConst.Organization.COLLECTION_NAME, null, null, 0, 0);
```
livthomas: Should not we return DTO here as well instead of providing the database entry structure to the public?

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
```diff
@@ -85,11 +85,10 @@ public void setCurrentProjectCode(final String projectCode) {
...
     * @param organizationCode
     *       code of the organization
-    * @return map(code, name) values of all projects in an organization
+    * @return list of all projects in an organization
     */
-   public Map<String, String> readProjectsMap(final String organizationCode) {
-      List<DataDocument> documents = dataStorage.search(Project.COLLECTION_NAME, organizationIdFilter(organizationFacade.getOrganizationId(organizationCode)), Arrays.asList(Project.ATTR_PROJECT_NAME, Project.ATTR_PROJECT_CODE));
-      return documents.stream().collect(Collectors.toMap(d -> d.getString(Project.ATTR_PROJECT_CODE), d -> d.getString(Project.ATTR_PROJECT_NAME)));
+   public List<DataDocument> readProjects(final String organizationCode) {
+      return dataStorage.search(Project.COLLECTION_NAME, organizationIdFilter(organizationFacade.getOrganizationId(organizationCode)), null, 0 ,0);
```
livthomas: The same problem with not using DTO here.

api/src/main/java/io/lumeer/engine/api/dto/Organization.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+package io.lumeer.engine.api.dto;
+
+import io.lumeer.engine.api.LumeerConst;
+import io.lumeer.engine.api.data.DataDocument;
+
+/**
+ * DTO object for Organization
+ */
+public class Organization {
```
livthomas: I think we can make this class `@Immutable` by removing setters and making all attributes `final` as we discussed.

api/src/main/java/io/lumeer/engine/api/dto/Project.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+package io.lumeer.engine.api.dto;
+
+import io.lumeer.engine.api.LumeerConst;
+import io.lumeer.engine.api.data.DataDocument;
+
+/**
+ * DTO object for Project
+ */
+public class Project {
```
livthomas: The same here. Let's make it `@Immutable`.

api/src/main/java/io/lumeer/engine/api/dto/Organization.java
livthomas: I would also make sense to implement `Serializable` interface here.

PR #67, opened 2017-06-04
Added new methods to organization and project services LMR-195

kubedo8:
I added new dependency to project api, so don't forget to make `mvn clean install` in *root* directory and then in *api* directory after merging this pull request.


api/pom.xml
```diff
@@ -60,6 +61,11 @@
             <type>pom</type>
             <scope>import</scope>
          </dependency>
+         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>${version.jackson.annotations}</version>
```
livthomas: This needs to be moved to `bom/pom.xml`.

PR #69, opened 2017-06-08
Created UserSettingsFacade and Service #LMR-204

kubedo8:
I'm not sure about scope of `Facade` and also `Service` but i'm too tired to think about it :D

### DIFF
--- a/api/src/main/java/io/lumeer/engine/api/dto/UserSettings.java
+++ b/api/src/main/java/io/lumeer/engine/api/dto/UserSettings.java
@@ -33,27 +33,20 @@ import javax.annotation.concurrent.Immutable;
 @Immutable
 public class UserSettings {
 
-   private final String user;
    private final String activeOrganization;
    private final String activeProject;
 
    public UserSettings(final DataDocument dataDocument) {
-      this(dataDocument.getString(LumeerConst.UserSettings.ATTR_USER),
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/lumeer/engine/api/dto/UserSettings.java
```diff
@@ -0,0 +1,72 @@
...
+
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * DTO object to store user settings
+ */
+@Immutable
+public class UserSettings {
+
+   private final String user;
```
livthomas: I do not think we need to transfer this information even though it is stored in the database. You can only get your settings and the application knows who you are. :)

war/src/main/java/io/lumeer/engine/rest/UserSettingsService.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Service for manipulation with user settings.
+ */
+@Path("/settings")
```
livthomas: I would recommend using `/settings/user` here and not use `user` parameter in any of those methods.

PR #70, opened 2017-06-09
Storing ids instead of codes in UserSettingsFacade #LMR-204

kubedo8:


war/src/main/java/io/lumeer/engine/controller/UserSettingsFacade.java
```diff
@@ -65,15 +81,27 @@ public UserSettings readUserSettings() {
     *       Dto object for user settings.
     **/
    public void upsertUserSettings(UserSettings userSettings) {
-      DataDocument dataDocument = userSettings.toDataDocument();
-      dataDocument.values().removeIf(Objects::isNull);
+      DataDocument dataDocument = new DataDocument();
+      if (userSettings.getDefaultOrganization() != null) {
+         String organizationId = organizationFacade.getOrganizationId(userSettings.getDefaultOrganization());
+         if (organizationId == null) {
+            return;
```
livthomas: Should not we throw exception in such cases? Because if `organizationId` is `null`, it means there is a problem.

kubedo8: Maybe we should..this situation means that organization doesn't exist.

PR #71, opened 2017-06-10
Improve collection metadata structure

alicak:
I changed attributes structure to flat after discussion with Tomáš. Some method got much simpler, but some will require more advanced queries, as we need to change values in documents which are stored inside an array. For now, they throw `UnsupportedOperationException`.


war/src/main/java/io/lumeer/engine/rest/dao/Attribute.java
```diff
@@ -19,50 +19,34 @@
...
 import io.lumeer.engine.api.data.DataDocument;
 
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
 /**
  * @author <a href="alica.kacengova@gmail.com">Alica Kačengová</a>
  */
 public class Attribute {
```
livthomas: Let's move both `Attribute` and `CollectionMetadata` classes to `io.lumeer.engine.api.dto` package where you can find examples of other similar DTOs.

war/src/main/java/io/lumeer/engine/rest/dao/Attribute.java
```diff
@@ -86,22 +74,22 @@ public boolean equals(final Object o) {
...
-   }
+   //   @Override
+   //   public String toString() {
+   //      return "Attribute{"
+   //            + ", name='" + name + '\''
+   //            + ", count=" + count
+   //            + ", constraints=" + constraints
+   //            + ", childAttributes=" + childAttributes
+   //            + '}';
+   //   }
```
livthomas: We should remove this.

war/src/main/java/io/lumeer/engine/rest/dao/Attribute.java
```diff
@@ -19,50 +19,34 @@
...
 
 /**
  * @author <a href="alica.kacengova@gmail.com">Alica Kačengová</a>
  */
 public class Attribute {
 
    private String name;
+   private String fullName;
    private int count;
+   private int level;
```
livthomas: What is this field good for? Do we really need it?

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -304,17 +293,16 @@ public void renameAttribute(String collectionName, String oldName, String newNam
...
-      String oldNamePath = attributePath(oldName);
-      String newNamePath = attributePath(newName);
+      //DataDocument renameQuery = dialect.renameAttributeQuery(metadataCollection(), collectionName, oldName, newName);
+      //dataStorage.run(renameQuery);
 
-      DataDocument renameQuery = dialect.renameAttributeQuery(metadataCollection(), collectionName, oldNamePath, newNamePath);
-
-      dataStorage.run(renameQuery);
+      // TODO
+      throw new UnsupportedOperationException();
```
livthomas: Yeah, this one will be hard. :)

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -333,6 +321,9 @@ public void dropAttribute(String collectionName, String attributeName) {
             dialect.concatFields(
                   LumeerConst.Collection.ATTRIBUTES_KEY,
                   attributePath(attributeName)));
+
+      throw new UnsupportedOperationException();
+      // TODO: drop children
```
livthomas: We just need to drop all attributes where the full name starts with the full name of the attribute which is being dropped. So probably a new filter similar to the wildcard filter will be needed.

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -354,36 +345,19 @@ public void addOrIncrementAttribute(String collectionName, String attributeName)
...
-         dataStorage.incrementAttributeValueBy(
-               metadataCollection(),
-               dialect.documentIdFilter(id),
-               dialect.concatFields(
-                     LumeerConst.Collection.ATTRIBUTES_KEY,
-                     attributePath(attributeName),
-                     LumeerConst.Collection.ATTRIBUTE_COUNT_KEY),
-               1);
+         throw new UnsupportedOperationException();
+         // TODO: use query
```
livthomas: Let's suppose `attributeName` is the full name here and update it accordingly. Or is there some problem I am not aware of?

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -414,14 +388,8 @@ public void dropOrDecrementAttribute(String collectionName, String attributeName
...
-      dataStorage.incrementAttributeValueBy(
-            metadataCollection(),
-            dialect.documentIdFilter(id),
-            dialect.concatFields(
-                  LumeerConst.Collection.ATTRIBUTES_KEY,
-                  attributePath(attributeName),
-                  LumeerConst.Collection.ATTRIBUTE_COUNT_KEY),
-            -1);
+      throw new UnsupportedOperationException();
+      // TODO: use query for count decrement
```
livthomas: The same as the previous one.

livthomas:
This work will be finished in #74.

PR #72, opened 2017-06-22
Enable more configuration options #LMR-213

kubedo8:


api/src/main/java/io/lumeer/engine/api/LumeerConst.java
```diff
@@ -243,4 +243,15 @@ public static LinkDirection fromString(final String s) {
       public static final String ATTR_GROUPS = "groups";
 
    }
+
+   public static class Configuration {
+
+      public static final String IDENTIFICATOR = "identificator";
```
livthomas: I would suggest to change this name since it might be confused with `_id`.

war/src/main/java/io/lumeer/engine/controller/ConfigurationFacade.java
```diff
@@ -42,14 +42,14 @@
 public class ConfigurationFacade implements Serializable {
 
    public enum ConfigurationLevel {
-      USER, PROJECT, ORGANISATION
+      USER, PROJECT, ORGANIZATION, GLOBAL
```
livthomas: This hierarchy is kind of weird. Why do not we have all five levels here?

- USER_GLOBAL
- USER_PROJECT
- USER_ORGANIZATION
- PROJECT
- ORGANIZATION

war/src/main/java/io/lumeer/engine/controller/configuration/ConfigurationManipulator.java
```diff
@@ -20,17 +20,21 @@
 package io.lumeer.engine.controller.configuration;
 
 import io.lumeer.engine.annotation.SystemDataStorage;
+import io.lumeer.engine.api.LumeerConst;
 import io.lumeer.engine.api.data.DataDocument;
 import io.lumeer.engine.api.data.DataFilter;
 import io.lumeer.engine.api.data.DataStorage;
 import io.lumeer.engine.api.data.DataStorageDialect;
+import io.lumeer.engine.api.dto.Config;
 import io.lumeer.engine.util.Resources;
```
livthomas: There are unused imports in this class.

war/src/main/java/io/lumeer/engine/controller/configuration/ConfigurationManipulator.java
```diff
@@ -65,251 +61,136 @@
...
 
-   public boolean hasConfigurationAttributeFlag(final String collectionName, final String configName, final String attributeName, final String flagName) {
-      Optional<DataDocument> configuration = getConfigurationEntry(collectionName, configName);
-      if (!configuration.isPresent()) {
-         return false;
-      }
-
-      List<String> flags = configuration.get().getArrayList(FLAGS_KEY, String.class);
-      if (flags.isEmpty()) {
+   public boolean hasConfigurationAttributeFlag(final String collectionName, final String identificator, final String attributeName, final String flagName) {
```
livthomas: I think we do not need this flag mechanism in `ConfigurationManipulator` anymore since the flags can be obtained directly from configuration entries.

PR #74, opened 2017-06-24
Improve collection metadata structure #LMR-199

kubedo8:


war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -219,8 +215,14 @@ public void createInitialMetadata(String internalCollectionName, String original
     * @return set of collection attributes' names
     */
    public Set<String> getAttributesNames(String collectionName) {
-      // TODO - use projection
-      return getAttributesInfo(collectionName).keySet();
+      String attributeNameKey = dialect.concatFields(Collection.ATTRIBUTES_KEY, Collection.ATTRIBUTE_FULL_NAME_KEY);
+      DataDocument metaData = readMetadata(collectionName, internalNameFilter(collectionName), Collections.singletonList(attributeNameKey));
+
+      return metaData != null ? metaData.getArrayList(Collection.ATTRIBUTES_KEY, DataDocument.class)
```
livthomas: This syntax may be very confusing. What about using something like this?
```java
if (metaData == null) {
   return Collections.emptySet();
}
return metaData.getArrayList...
```

PR #75, opened 2017-06-27
Implementation of the SecurityService #LMR-203

martinglova:
Tha last pulls had to do something with URLs for my tests. I think there could be conflicts.

I've defined URLs as in the updated comment: https://lumeer.myjetbrains.com/youtrack/issue/LMR-203#comment=100-114

But I temporarily added `/roles` before all of the URLs in the code to pass all the tests.

### DIFF
--- a/war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
+++ b/war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
@@ -25,13 +25,16 @@ import io.lumeer.engine.api.LumeerConst;
 import io.lumeer.engine.api.data.DataDocument;
 import io.lumeer.engine.api.data.DataStorage;
 import io.lumeer.engine.api.data.DataStorageDialect;
+import io.lumeer.engine.api.dto.Role;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.zip.DataFormatException;
 import javax.enterprise.context.SessionScoped;
 import javax.inject.Inject;
 
@@ -341,5 +344,51 @@ public class SecurityFacade implements Serializable {
 
    private String userOrGroupName(String user, String group) {
       return user != null ? user : group;
+   }
+   /********* GETTING ROLES *********/
+   public List<Role> getOrganizationRoles(String organizationCode) {
+      DataDocument doc = systemDataStorage.readDocumentIncludeAttrs(
+            LumeerConst.Security.ORGANIZATION_ROLES_COLLECTION_NAME,
+            dataStorageDialect.fieldValueFilter(LumeerConst.Security.ORGANIZATION_ID_KEY,
+                  organizationFacade.getOrganizationId(organizationCode)),
+            Arrays.asList(LumeerConst.Security.ROLES_KEY));
+
+      return getRolesList(doc);
+   }
+
+   public List<Role> getProjectRoles(String projectCode) {
+      DataDocument doc = dataStorage.readDocumentIncludeAttrs(
+            LumeerConst.Security.ROLES_COLLECTION_NAME,
+            dataStorageDialect.multipleFieldsValueFilter(projectFilter(projectCode)),
+            Arrays.asList(dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY)));
+
+      return getRolesList(doc);
+   }
+
+   public List<Role> getCollectionRoles(String projectCode, String collectionName) {
+      DataDocument doc = dataStorage.readDocumentIncludeAttrs(
+            LumeerConst.Security.ROLES_COLLECTION_NAME,
+            dataStorageDialect.multipleFieldsValueFilter(collectionFilter(projectCode, collectionName)),
+            Arrays.asList(dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY)));
+
+      return getRolesList(doc);
+   }
+
+   public List<Role> getViewRoles(String projectCode, int viewId) {
+      DataDocument doc = dataStorage.readDocumentIncludeAttrs(
+            LumeerConst.Security.ROLES_COLLECTION_NAME,
+            dataStorageDialect.multipleFieldsValueFilter(viewFilter(projectCode, viewId)),
+            Arrays.asList(dataStorageDialect.concatFields(LumeerConst.Security.ROLES_KEY)));
+
+      return getRolesList(doc);
+   }
+
+   private List<Role> getRolesList(DataDocument doc) {
+      List<Role> output = new ArrayList<>();
+      doc.getDataDocument(LumeerConst.Security.ROLES_KEY).forEach((k, v)-> {
+         output.add(new Role(k, (DataDocument) v));
+      });
+
+      return output;
    }
 }
--- a/war/src/main/java/io/lumeer/engine/rest/SecurityService.java
+++ b/war/src/main/java/io/lumeer/engine/rest/SecurityService.java
@@ -1,0 +1,425 @@
+/*
+ * -----------------------------------------------------------------------\
+ * Lumeer
+ *  
+ * Copyright (C) since 2016 the original author or authors.
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------/
+ */
+package io.lumeer.engine.rest;
+
+import io.lumeer.engine.api.LumeerConst;
+import io.lumeer.engine.api.dto.Role;
+import io.lumeer.engine.controller.OrganizationFacade;
+import io.lumeer.engine.controller.ProjectFacade;
+import io.lumeer.engine.controller.SecurityFacade;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.zip.DataFormatException;
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.GET;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
+
+@Path("/roles/organizations/{organization}")
+@RequestScoped
+@Produces(MediaType.APPLICATION_JSON)
+public class SecurityService implements Serializable {
+
+   @PathParam("organization")
+   private String organizationCode;
+
+   @Inject
+   private SecurityFacade securityFacade;
+   @Inject
+   private OrganizationFacade organizationFacade;
+   @Inject
+   private ProjectFacade projectFacade;
+
+   @PostConstruct
+   public void init() {
+      if (organizationCode == null)
+         throw new BadRequestException();
+
+      organizationFacade.setOrganizationCode(organizationCode);
+   }
+
+   private boolean isRoleCorrect(String resource, String roleName) {
+      return LumeerConst.Security.RESOURCE_ROLES.get(resource).contains(roleName);
+   }
+
+   /**
+    * Add a role for users and groups for an organization.
+    *
+    * @param role
+    *       the role to be added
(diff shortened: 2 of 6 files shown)


api/src/main/java/io/lumeer/engine/api/dto/Roles.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.zip.DataFormatException;
+/**
+ * @author <a href="mailto:martin.glova@lumeer.io">Martin Glova</a>
+ */
+public class Roles extends DataDocument {
+   public Roles(final DataDocument data, final String resource) throws DataFormatException {
```
kubedo8: DTO objects should look like others. So it have to contains string,integer and other needed fields..

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
```diff
@@ -127,7 +128,7 @@ public Organization readOrganization(final String organizationCode) {
     *       organization data
     * @return id of the organization
     */
-   public String createOrganization(final Organization organization) {
+   public String createOrganization(final Organization organization) throws InvalidDocumentKeyException {
       String id = dataStorage.createDocument(LumeerConst.Organization.COLLECTION_NAME, organization.toDataDocument());
       databaseInitializer.onOrganizationCreated(id);
       securityFacade.addOrganizationUserRole(organization.getCode(), userFacade.getUserEmail(), LumeerConst.Security.ROLE_MANAGE);
```
kubedo8: Move this line to `databaseInitilizer.onOrganizationCreated` method and don't throw exception because we know, that we add correct role.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
```diff
@@ -147,7 +148,7 @@ public Project readProject(final String projectCode) {
     *       project data
     * @return id of the organization
     */
-   public String createProject(final Project project) {
+   public String createProject(final Project project) throws InvalidDocumentKeyException {
       DataDocument dataDocument = project.toDataDocument()
                                          .append(LumeerConst.Project.ATTR_ORGANIZATION_ID, organizationFacade.getOrganizationId());
       String id = dataStorage.createDocument(LumeerConst.Project.COLLECTION_NAME, dataDocument);
```
kubedo8: Same here. Move adding role to `databaseInitilizer.onOrganizationCreated` method and don't throw exception because we know, that we add correct role.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
```diff
@@ -83,7 +84,7 @@ public int createView(
          final String originalViewName,
          final String viewType,
          final String description,
-         final DataDocument configuration) throws ViewAlreadyExistsException {
+         final DataDocument configuration) throws ViewAlreadyExistsException, InvalidDocumentKeyException {
```
kubedo8: Same here...

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
martinglova: What do you mean by move this line? It is a line with the declaration of a method.

But finally, I see that data are not dependent on input so I suggest to do:

`try {`
`   securityFacade.addOrganizationUserRole(organization.getCode(), userFacade.getUserEmail(), LumeerConst.Security.ROLE_MANAGE);`
`} catch (InvalidDocumentKeyException e) {`
`   // Never can happen`
`}`

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
martinglova: Same here...

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
martinglova: Same here...

war/src/main/java/io/lumeer/engine/controller/SecurityFacade.java
```diff
@@ -342,4 +345,53 @@ private String userOrGroupKey(String user) {
...
+      return getRolesList(doc);
+   }
+
+   private List<Role> getRolesList(DataDocument doc) {
+      List<Role> output = new ArrayList<>();
+      doc.getDataDocument(LumeerConst.Security.ROLES_KEY).forEach((k, v)-> {
+         output.add(new Role(
+               k,
+               ((Map<String, List<String>>) v).get(LumeerConst.Security.USERS_KEY),
+               ((Map<String, List<String>>) v).get(LumeerConst.Security.GROUP_KEY)));
```
livthomas: `Role` class should have a constructor that takes `DataDocument` as a parameter in order to make this method simpler.

war/src/test/java/io/lumeer/engine/controller/SecurityFacadeIntegrationTest.java
```diff
@@ -106,8 +109,24 @@ public void addAndRemoveOrganizationUserRole() throws Exception {
       securityFacade.addOrganizationUserRole(org, user, roleManage);
       assertThat(securityFacade.hasOrganizationRole(org, roleManage)).isTrue();
 
+      List<Role> roles = securityFacade.getOrganizationRoles(org);
+      roles.forEach((item) -> {
+         assertThat(LumeerConst.Security.RESOURCE_ROLES.get(LumeerConst.Security.ORGANIZATION_RESOURCE).contains(item.getName())).isTrue();
+         if (item.getName().equals(roleManage)) {
+            assertThat(item.getUsers().contains(user)).isTrue();
```
livthomas: This can be made much simpler:
```java
assertThat(roles).extracting(role -> role.getName())
 .containsOnly(LumeerConst.Security.RESOURCE_ROLES.get(LumeerConst.Security.ORGANIZATION_RESOURCE));

assertThat(roles).filteredOn(role -> role.getName().equals(roleManage))
 .extracting(role -> role.getUsers().contains(user);
```

The same applies to other such blocks.

war/src/test/java/io/lumeer/engine/rest/LinkingServiceIntegrationTest.java
```diff
@@ -89,7 +90,7 @@
    private UserFacade userFacade;
 
    @Before
-   public void init() {
+   public void init() throws InvalidDocumentKeyException {
```
livthomas: Why is this needed now and was not needed before?

war/src/test/java/io/lumeer/engine/rest/SecurityServiceIntegrationTest.java
```diff
@@ -0,0 +1,1413 @@
...
+            .path(createUrl(
+                  LumeerConst.Security.ORGANIZATION_RESOURCE,
+                  organizationCode,
+                  LumeerConst.Security.ROLES_KEY,
+                  role))
+            .queryParam(USERS_QP, USER_X, USER_Y)
+            .queryParam(GROUPS_QP, GROUP_A, GROUP_B)
+            .request().build(PUT_METHOD_NAME)
+            .invoke();*/
+      // TODO: Checking throwing exception doesn't work.
```
livthomas: Check HTTP codes as we discussed.

war/src/test/java/io/lumeer/engine/rest/SecurityServiceIntegrationTest.java
```diff
@@ -0,0 +1,1413 @@
...
+            .target(TARGET_URI)
+            .path(createUrl(
+                  LumeerConst.Security.ORGANIZATION_RESOURCE,
+                  organizationCode,
+                  LumeerConst.Security.ROLES_KEY))
+            .request().buildGet()
+            .invoke();
+
+      List<Role> fromService = response.readEntity(List.class);
+      assertThat(fromFacade.size()).isEqualTo(fromService.size());
```
livthomas: Please, use collection assertions. It is easier to see where the problem is when such assertion fails.
```java
assertThat(fromService).hasSize(fromFacade.size());
```

war/src/test/java/io/lumeer/engine/rest/SecurityServiceIntegrationTest.java
```diff
@@ -0,0 +1,1413 @@
...
+      for (Role r : fromFacade) {
+         assertThat(fromService.contains(r));
```
livthomas: Make better use of AssertJ:
```java
assertThat(fromService).containsExactly(fromFacade);
```

war/src/test/java/io/lumeer/engine/rest/SecurityServiceIntegrationTest.java
```diff
@@ -0,0 +1,1413 @@
...
+      List<String> groups = null;
+      for (Role item: fromFacade) {
+         if (item.getName().equals(role)) {
+            users = item.getUsers();
+            groups = item.getGroups();
+         }
+      }
+
+      assertThat(users.contains(USER_X)).isTrue();
+      assertThat(users.contains(USER_Y)).isFalse();
```
livthomas:
```java
assertThat(users).containsOnly(USER_X);
```

war/src/main/java/io/lumeer/engine/rest/SecurityService.java
```diff
@@ -0,0 +1,410 @@
...
+         final @QueryParam("groups") List<String> groups) {
+      if (role == null || !isRoleCorrect(LumeerConst.Security.ORGANIZATION_RESOURCE, role))
+         throw new BadRequestException();
+
+      for (String user: users) {
+         securityFacade.addOrganizationUserRole(organizationCode, user, role);
+      }
+
+      for (String group: groups) {
+         securityFacade.addOrganizationGroupRole(organizationCode, group, role);
```
livthomas: This is highly inefficient since for every user and group we make a new database call. Consider adding appropriate method to the facade to handle such operations.

war/src/main/java/io/lumeer/engine/rest/SecurityService.java
```diff
@@ -0,0 +1,410 @@
...
+    * Get roles of an organization.
+    *
+    * @return
+    *       the role object with all roles for the organization
+    * @throws DataFormatException
+    *       if obtained data cannot be transformed to a role object
+    */
+   @GET
+   @Path("/roles")
+   @Produces(MediaType.APPLICATION_JSON)
```
livthomas: Move all `@Produces` and `@Consumes` annotations to the class level.

war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java
martinglova: Will do for the next commit.

war/src/main/java/io/lumeer/engine/controller/ProjectFacade.java
martinglova: Will do for the next commit.

war/src/main/java/io/lumeer/engine/controller/ViewFacade.java
martinglova: Will do for the next commit.

war/src/main/java/io/lumeer/engine/rest/SecurityService.java
```diff
@@ -0,0 +1,405 @@
...
+    *       the project code to be added the role for
+    * @param role
+    *       the role to be added
+    * @param users
+    *       the users to be added the role for
+    * @param groups
+    *       the groups to be added the role for
+    */
+   @PUT
+   @Path("/projects/{project}/roles/{role}")
```
livthomas: You should call `projectFacade.setProjectCode(projectCode)` after the initial check in this method as well as all other methods with `project` parameter.

war/src/test/java/io/lumeer/engine/controller/SecurityFacadeIntegrationTest.java
```diff
@@ -106,8 +108,27 @@ public void addAndRemoveOrganizationUserRole() throws Exception {
...
       assertThat(securityFacade.hasOrganizationRole(org, roleManage)).isTrue();
 
+      List<Role> roles = securityFacade.getOrganizationRoles(org);
+      assertThat(roles).extracting(role -> role.getName())
+                       .containsOnly(LumeerConst.Security.RESOURCE_ROLES.get(
+                             LumeerConst.Security.ORGANIZATION_RESOURCE).toArray(new String[LumeerConst.Security.RESOURCE_ROLES.get(
+                                   LumeerConst.Security.ORGANIZATION_RESOURCE).size()]));
+
+      assertThat(roles).filteredOn(role -> role.getName().equals(roleManage))
+                       .extracting(role -> role.getUsers().contains(user));
```
livthomas: `extracting()` method itself does not check anything. You need to call `extracting(role -> role.getUsers()).contains(user)`.

war/src/main/java/io/lumeer/engine/rest/SecurityService.java
kubedo8: @livthomas I think that it's not necessary in this case, isn't it?

livthomas: @kubedo8 probably not but we should rather set it since we rely on this in many parts of our application. I plan to change this whole mechanism around organization and project context in the future.

PR #77, opened 2017-07-04
Improve attribute search suggestions #LMR-222

kubedo8:


api/src/main/java/io/lumeer/engine/api/dto/SearchSuggestion.java
```diff
@@ -46,7 +49,9 @@ public SearchSuggestion(final String type, final String text, final String icon)
       this(type, text, Collections.emptyList(), icon);
    }
 
-   public SearchSuggestion(final String type, final String text, final List<String> constraints, final String icon) {
+   @JsonCreator
+   public SearchSuggestion(final @JsonProperty("type") String type, final @JsonProperty("text") String text,
+         final @JsonProperty("constraints") List<String> constraints, final @JsonProperty("icon") String icon) {
```
livthomas: Do we really need this? I think we are sending suggestions just out.

war/src/test/java/io/lumeer/engine/controller/search/QuerySuggesterIntegrationTest.java
```diff
@@ -188,6 +183,22 @@ public void testSuggestAttributesCompleteAttribute() throws Exception {
...
+   public void testSuggestAttributesChilds() throws Exception {
+      collectionFacade.createCollection(COLLECTION_COMPANIES);
+      String internalCollectionName = collectionMetadataFacade.getInternalCollectionName(COLLECTION_COMPANIES);
+      collectionMetadataFacade.addOrIncrementAttribute(internalCollectionName, "Lumeer");
+      collectionMetadataFacade.addOrIncrementAttribute(internalCollectionName, "First.Lumeer");
+      collectionMetadataFacade.addOrIncrementAttribute(internalCollectionName, "Second.First.Lumeer");
+      collectionMetadataFacade.addOrIncrementAttribute(internalCollectionName, "a.b.c.d.e.Lumee");
+
+      List<SearchSuggestion> attributes = querySuggester.suggestAttributes("companies.lumee", QuerySuggester.SUGGESTIONS_LIMIT);
+      assertThat(attributes).hasSize(4);
```
livthomas: What about checking the structure of returned data as well?

war/src/main/java/io/lumeer/engine/controller/search/QuerySuggester.java
```diff
@@ -105,41 +104,56 @@
...
-                                  .collect(Collectors.toList()); // TODO do not ignore child attributes
+         return filterAttributes(collectionMetadata, attributePart, limit);
       } catch (UserCollectionNotFoundException e) {
          return Collections.emptyList();
       }
    }
 
+   private List<SearchSuggestion> suggestAttributesForAllCollections(String attributePart, int limit) {
+      String metadataCollection = collectionMetadataFacade.metadataCollection();
+      List<DataDocument> collectionsRaw = userDataStorage.search(metadataCollection, null, null);
```
livthomas: Isn't it possible to get attributes directly instead of looping through all metadata collections?

api/src/main/java/io/lumeer/engine/api/dto/SearchSuggestion.java
kubedo8: Yes because some test throws `nullpointerexception` after changing `response.readEntity(List.class);` to `response.readEntity(new GenericType<List<SearchSuggestion>>(){});` which is (my) preffered way to read data in rest tests and it doesn't show warning..

war/src/main/java/io/lumeer/engine/controller/search/QuerySuggester.java
kubedo8: I will try it

war/src/test/java/io/lumeer/engine/controller/search/QuerySuggesterIntegrationTest.java
kubedo8: Hmmm I think that it's not necessary, because structure is checking in other tests, but I can do that to make sure everything works fine

PR #78, opened 2017-07-07
Fix resource default permissions #LMR-233

livthomas:
I have changed the default permissions so the user who created a resource is able to do anything with it by default.


kubedo8:
Some tests are failing..

PR #79, opened 2017-07-08
Edit CollectionService for Rest purpose #LMR-235

kubedo8:
This pull request is initially not fully implemented and also not tested. I want to discuss some things with @livthomas to be sure I'm on right way.


war/src/main/java/io/lumeer/engine/rest/CollectionService.java
```diff
@@ -121,38 +118,71 @@ public void init() {
     */
    @GET
    @Path("/")
-   public List<String> getAllCollections() {
-      List<String> collections = new ArrayList<>();
+   public List<Collection> getCollections(@QueryParam("collection") String collectionNamePart, @QueryParam("limit") int limit) {
```
livthomas: Why do we need to query collections by partial name? `SearchService` is responsible for such queries or am I missing something?

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -184,26 +182,71 @@ public String createInternalName(String originalCollectionName) throws UserColle
       return name;
    }
 
+   public List<io.lumeer.engine.api.dto.Collection> getCollectionsMetadata(String collectionNamePart) {
```
livthomas: Do not we want to import `Collection` class?

api/src/main/java/io/lumeer/engine/api/dto/Collection.java
```diff
@@ -0,0 +1,83 @@
...
+            document.getString(LumeerConst.Collection.ICON),
+            document.getString(LumeerConst.Collection.COLOR),
+            document.getInteger(LumeerConst.Collection.DOCUMENT_COUNT));
+   }
+
+   @JsonCreator
+   public Collection(final @JsonProperty(LumeerConst.Collection.REAL_NAME) String name,
+         final @JsonProperty(LumeerConst.Collection.ICON) String icon,
+         final @JsonProperty(LumeerConst.Collection.COLOR) String color,
+         final @JsonProperty(LumeerConst.Collection.DOCUMENT_COUNT) int docCount) {
```
kubedo8: Yes that's true..

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
kubedo8: Because there is import of `LumeerConst.Collection` and it causes conflict, but I will refactor this import to be more cleaner what is going on...

war/src/main/java/io/lumeer/engine/controller/BatchFacade.java
```diff
@@ -154,18 +157,19 @@ private void internalExecuteBatch(final MergeBatch batch) throws DbException, In
             doc.put(batch.getResultAttribute(), subDoc);
          }
 
+         String collectionCode = collectionMetadataFacade.getCollectionCode(batch.getCollectionName());
```
livthomas: I think there should be `collectionCode` stored in `Batch` objects. But it is probably outside of the scope of this pull-request.

war/src/main/java/io/lumeer/engine/controller/CollectionFacade.java
```diff
@@ -90,309 +78,226 @@
...
     * @throws AttributeAlreadyExistsException
     *       when attribute with new name already exists
     */
-   public void renameAttribute(final String collectionName, final String origName, final String newName) throws CollectionNotFoundException, AttributeAlreadyExistsException {
-      if (dataStorage.hasCollection(collectionName)) {
-         collectionMetadataFacade.renameAttribute(collectionName, origName, newName);
-         dataStorage.renameAttribute(collectionName, origName, newName);
+   public void renameAttribute(final String collectionCode, final String collectionName, final String origName, final String newName) throws AttributeAlreadyExistsException {
+      collectionMetadataFacade.renameAttribute(collectionCode, origName, newName);
+      dataStorage.renameAttribute(collectionName, origName, newName);
```
livthomas: Why do we pass `collectionName` here? Should not collections be identified by `collectionCode` only?

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -138,205 +128,268 @@ public DataDocument getCollectionMetadataDocument(String collectionName) throws
...
-      name = Collection.NAME_PREFIX + name;
-      int i = 0;
-      while (dataStorage.hasCollection(name + "_" + i)) {
-         i++;
-      }
-      name = name + "_" + i;
+   /**
+    * @return map of collection codes and their names
+    */
+   public Map<String, String> getCollectionsCodeName() {
```
livthomas: Do we really need this method?

war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
```diff
@@ -235,14 +235,14 @@ private void initProjectRoles(String projectId) {
...
+    * @param collectionCode
+    *       collectionCode
     */
-   private void initCollectionRoles(String projectCode, String collectionName) {
+   private void initCollectionRoles(String projectCode, String collectionCode) {
       DataDocument roles = new DataDocument()
             .append(Security.PROJECT_ID_KEY, projectFacade.getProjectId(projectCode))
             .append(Security.TYPE_KEY, Security.TYPE_COLLECTION)
-            .append(Security.COLLECTION_NAME_KEY, collectionName)
+            .append(Security.COLLECTION_CODE_KEY, collectionCode)
```
livthomas: Should not we use collection ID here?

war/src/main/java/io/lumeer/engine/controller/DocumentFacade.java
```diff
@@ -81,13 +81,15 @@
...
+    * @param collectionCode
+    *       collection code
     * @param document
     *       the DataDocument object representing a document to be created
     * @return the id of the newly created document
     * @throws DbException
     *       When there is an error working with the database.
     */
-   public String createDocument(final String collectionName, final DataDocument document) throws DbException {
+   public String createDocument(final String collectionName, final String collectionCode, final DataDocument document) throws DbException {
```
livthomas: Why do we need `collectionName` parameter here? It does not uniquely identify collections.

war/src/main/java/io/lumeer/engine/rest/CollectionService.java
```diff
@@ -121,70 +116,103 @@ public void init() {
     */
    @GET
    @Path("/")
-   public List<String> getAllCollections() {
-      List<String> collections = new ArrayList<>();
+   public List<Collection> getCollections(@QueryParam("limit") int limit) {
```
livthomas: I think we need proper paging here, not only `limit`. I would suggest using `page` and `size` parameters.

war/src/main/java/io/lumeer/engine/controller/DatabaseInitializer.java
kubedo8: `collectionCode` is also immutable so what is better?

war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
kubedo8: In some cases it's really helpful to return this two information like this.

PR #82, opened 2017-07-25
Implement organization permissions tests #LMR-250

SquattingPenguin:


war/src/test/java/io/lumeer/engine/rest/permissions/OrganizationServicePermissionsIntegrationTest.java
```diff
@@ -19,50 +19,322 @@
...
+                                .path(pathPrefix(code))
+                                .request()
+                                .buildPost(Entity.json(new Project(project, projectName)))
+                                .invoke();
+      String reasonPhrase = response.getStatusInfo().getReasonPhrase();
+      assertThat(reasonPhrase).isEqualTo("No Content");
+      client.close();
    }
 
+   private String pathPrefix(String organizationCode) {
```
livthomas: I would suggest renaming this to `projectPathPrefix` so the name of the method clearly says what it does.

PR #83, opened 2017-07-26
Fix adding documents with same attributes

justmesr:


war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -570,9 +570,17 @@ private Object checkAndConvertAttributeValue(final Attribute attribute, final Ob
...
+      // no encoding required
+      if (value == null) {
+         return null;
+      }
+
+      // other types
       final List<String> constraintConfigurations = attribute.getConstraints();
 
       final ConstraintManager constraintManager = new ConstraintManager(constraintConfigurations);
+      constraintManager.setLocale(Locale.getDefault());
```
livthomas: Why not using this locale by default in `ConstraintManager`?

justmesr: `ConstraintManager` almost never uses the locale, and objects using it might have more information about user Locale.

livthomas: Ok, sounds reasonable.

PR #84, opened 2017-07-27
Ignore unknown fields from rest

kubedo8:


livthomas:
I'm closing this since we have discussed that we should not transfer data that are not needed.

PR #85, opened 2017-07-27
Add get methods for documents and collections

justmesr:


war/src/main/java/io/lumeer/engine/rest/CollectionService.java
```diff
@@ -217,6 +217,38 @@ public void dropCollection(final @PathParam("collectionCode") String collectionC
...
+      if (!securityFacade.hasCollectionRole(projectCode, collectionCode, LumeerConst.Security.ROLE_READ)) {
+         throw new UnauthorizedAccessException();
+      }
+
+      return collectionFacade.getCollections().stream()
+                             .filter(collection -> securityFacade.hasCollectionRole(projectCode, collection.getCode(), LumeerConst.Security.ROLE_READ))
+                             .filter(collection -> collection.getCode().equals(collectionCode))
+                             .findFirst()
+                             .orElseThrow(() -> new CollectionNotFoundException(ErrorMessageBuilder.userCollectionNotFoundString(collectionCode)));
+   }
```
kubedo8: You should add new method to `CollectionFacade` to read collection directly, because reading all collections is very unefficient.

war/src/main/java/io/lumeer/engine/controller/CollectionFacade.java
```diff
@@ -309,7 +309,7 @@ public void dropAttributeConstraint(final String collectionCode, final String at
     *       name of the collection
     * @return list of all documents
     */
-   private List<DataDocument> getAllDocuments(String collectionCode) {
+   public List<DataDocument> getAllDocuments(String collectionCode) {
```
livthomas: This method should rather be provided by `DocumentFacade`.

PR #89, opened 2017-08-03
Move collection security to collection metadata #LMR-287

kubedo8:


api/src/main/java/io/lumeer/engine/api/dto/Permission.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class Permission {
+
+   private final String name;
+   private final List<String> roles;
```
livthomas: This should be `Set` in order to prevent having the same role there twice.

api/src/main/java/io/lumeer/engine/api/dto/Permission.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class Permission {
```
livthomas: This class should have `equals` method based purely on `name` attribute.

war/src/main/java/io/lumeer/engine/rest/CollectionService.java
```diff
@@ -144,17 +148,17 @@ public void init() {
     * @return Collection with provided code
     */
    @GET
-   @Path("/{collectionCode}/")
+   @Path("/{collectionCode}")
```
livthomas: Just for the record, `/` before the parameter is not needed either. :)

war/src/main/java/io/lumeer/engine/rest/CollectionService.java
```diff
@@ -576,11 +580,141 @@ public void dropAttributeConstraint(final @PathParam("collectionCode") String co
...
+   public Map<String, List<Permission>> getPermissions(final @PathParam("collectionCode") String collectionCode) throws CollectionNotFoundException, UnauthorizedAccessException {
+      if (collectionCode == null) {
+         throw new BadRequestException();
+      }
+
+      if (!collectionFacade.hasCollection(collectionCode)) {
+         throw new UserCollectionNotFoundException(ErrorMessageBuilder.userCollectionNotFoundString(collectionCode));
+      }
+
+      if (!collectionMetadataFacade.hasRole(projectCode, collectionCode, LumeerConst.Security.ROLE_READ)) {
```
livthomas: Users with `read` role should not be able to see collection permissions. You need to have `manage` role in order to do so.

war/src/main/java/io/lumeer/engine/rest/CollectionService.java
```diff
@@ -144,17 +148,17 @@ public void init() {
...
       }
 
-      if (!collectionFacade.hasCollection(collectionCode)) {
+      Collection collection = collectionFacade.getCollection(collectionCode);
+      if (collection == null) {
          throw new UserCollectionNotFoundException(ErrorMessageBuilder.userCollectionNotFoundString(collectionCode));
       }
-
-      if (!securityFacade.hasCollectionRole(projectCode, collectionCode, LumeerConst.Security.ROLE_READ)) {
+      if (!collection.getUserRoles().contains(LumeerConst.Security.ROLE_MANAGE)) {
```
livthomas: Why does user need to have `manage` role in order to be able to read collection?

PR #90, opened 2017-08-06
Fix removing collection attributes LMR-291

justmesr:


war/src/main/java/io/lumeer/engine/controller/CollectionMetadataFacade.java
```diff
@@ -342,16 +344,24 @@ public void renameAttribute(String collectionCode, String oldFullName, String ne
...
+      if (collectionToRemoveFrom == null) {
          return;
       }
 
-      dataStorage.removeItemsFromArray(metadataCollection(), collectionCodeFilter(collectionCode), LumeerConst.Collection.ATTRIBUTES,
-            removeAttributes.getArrayList(LumeerConst.Collection.ATTRIBUTES, DataDocument.class));
+      // match "<attributeName>", or "<attributeName>.<child>"
+      Pattern pattern = Pattern.compile(Pattern.quote(attributeName) + "(\\..*)?");
+      Matcher matcher = pattern.matcher("");
+
```
kubedo8: And what about dropping attribute `(<something>)*.<attributeName>.<child>` ? Old code considered also this situations, but I think that now it's not working like that.  Or am I wrong?

kubedo8:
What improvement make this code? I'm not 100% sure, but I think that it does the same as it was before. (Or not?)

justmesr:
When dropping any attribute from collection, the method `dropAttribute` would drop all attributes before.

(In the code the `removeAttributes = dataStorage.readDocumentIncludeAttrs` fetched all attributes from the collection and then dropped them all.)

kubedo8:
Oh now I see, that previous filter can't work.
Altought there is unnecessarry reading attributes from db, it may remain as it is, because with current structure is not easy to implement appropriate request, but in the future we will do it more efficiently.

PR #91, opened 2017-08-07
 Fix first document in new collection not counted bug #LMR-290

SquattingPenguin:


api/src/main/java/io/lumeer/engine/api/data/DataStorage.java
```diff
@@ -111,13 +111,13 @@ default void connect(final StorageConnection connection, final String database,
...
-    * @param collectionName
-    *       the name of the collection where the document will be created
+    * @param collectionCode
+    *       the code of the collection where the document will be created
     * @param document
     *       the DataDocument object representing a document to be created
     * @return the id of the newly created document
     */
-   String createDocument(final String collectionName, final DataDocument document);
+   String createDocument(final String collectionCode, final DataDocument document);
```
livthomas: Why this change? This is not about collections in our terminology but MongoDB collections which have names.

justmesr:
>"Do we really want to enable users to create documents in collections which do not exist?"
Does it make sense to have empty collections? Empty tables in Excel are totally useless.
And the only difference I see with our case are the permissions (someone with manage role would have to create at least one document to have a collection)..
On the other hand it could simplify things on the frontend and CollectionService side. :)

api/src/main/java/io/lumeer/engine/api/data/DataStorage.java
```diff
@@ -112,7 +112,7 @@ default void connect(final StorageConnection connection, final String database,
     * Creates and inserts a new document to specified collection.
     *
     * @param collectionName
-    *       the name of the collection where the document will be created
+    *       the code of the collection where the document will be created
```
livthomas: This should also be reverted.

PR #92, opened 2017-08-10
Refactor views implementation #LMR-263

livthomas:
The new back-end structure demonstrated on the implementation of views stack.


lumeer-core/src/main/java/io/lumeer/core/AuthenticatedUser.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Set;
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+import javax.servlet.http.HttpServletRequest;
+
+@RequestScoped
+public class AuthenticatedUser {
+
+   public static final String DEFAULT_USERNAME = "demouser";
+   public static final String DEFAULT_EMAIL = "demouser@lumeer.io";
```
marvec: Please, make it more fancy. Like Alan Turing ;-)

lumeer-core/src/main/java/io/lumeer/core/WorkspaceKeeper.java
```diff
@@ -0,0 +1,66 @@
...
+   }
+
+   @Override
+   public Optional<Project> getProject() {
+      return Optional.ofNullable(project);
+   }
+
+   public void setOrganization(String organizationCode) {
+      this.organization = organizationDao.getOrganizationByCode(organizationCode); // TODO cache this in the future
+   }
```
marvec: Previous implementation was fast because it was purely in memory. I would suggest that we implement this cache NOW. Because this is the place through which every single request passes. It can easily become a bottleneck.

lumeer-core/src/main/java/io/lumeer/core/WorkspaceKeeper.java
```diff
@@ -0,0 +1,66 @@
...
+
+   public void setProject(String projectCode) {
+      this.project = projectDao.getProjectByCode(projectCode); // TODO cache this in the future
```
marvec: The same applies here.

lumeer-core/src/main/java/io/lumeer/core/facade/ViewFacade.java
```diff
@@ -0,0 +1,173 @@
...
+import io.lumeer.storage.api.dao.ViewDao;
+
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+
+@RequestScoped
```
marvec: What prevents us from making is Session Scoped?

lumeer-remote/src/main/java/io/lumeer/remote/rest/AbstractService.java
```diff
@@ -0,0 +1,50 @@
...
+   @Inject
+   private HttpServletRequest request;
+
+   protected URI getResourceURI(Resource resource) {
+      return UriBuilder.fromUri(request.getRequestURL() + "/" + resource.getCode()).build();
+   }
+
+   protected URI getParentURI() {
+      String fullPath = request.getRequestURL().toString();
+      String parentPath = fullPath.substring(0, fullPath.lastIndexOf("/"));
```
marvec: I am not sure that String manipulation is the smartest way of getting URI parts. I would prefer to rely on URI methods. There can be escaped slashes or some special occurrences to which this method does not apply. Like www.site.com/abc?p=a/b/c

lumeer-remote/src/main/java/io/lumeer/remote/rest/AbstractService.java
```diff
@@ -0,0 +1,50 @@
...
+      return UriBuilder.fromUri(parentPath).build();
+   }
+
+   protected URI getGrandParentURI() {
```
marvec: How about getParentService() method on which the getParentUri will be called? Also the method name violates our code convention - URI should be Uri. Or do we change the convention? From my experience, this convention is Useful for cases like getMariaDbDs vs get MariaDBDS - you cannot differentiate individual acronyms.

marvec:
It is a nice rework. I hope to see at least ten times bigger commit in web-ui soon! :-D

lumeer-core/src/main/java/io/lumeer/core/AuthenticatedUser.java
livthomas: Ok, I will use that.

lumeer-core/src/main/java/io/lumeer/core/WorkspaceKeeper.java
livthomas: This is not completely true. [OrganizationFacade](https://github.com/Lumeer/engine/blob/devel/war/src/main/java/io/lumeer/engine/controller/OrganizationFacade.java#L96-L99) made database requests in every request where `organizationCode` was provided.

lumeer-core/src/main/java/io/lumeer/core/facade/ViewFacade.java
livthomas: Because we need to initialize `ViewDao` which depends on the project specified in the request. Otherwise, it would be `ApplicationScoped`.

In the future, it would be better to introduce `ProjectScoped` and `OrganizationScoped` in order not to have too many instances of some beans which are not really needed.

lumeer-remote/src/main/java/io/lumeer/remote/rest/AbstractService.java
livthomas: `getRequestURL()` unlike `getRequestURI()` returns only the URL without query parameters so this is not a problem. However, I will rewrite it using regular expression.

lumeer-remote/src/main/java/io/lumeer/remote/rest/AbstractService.java
livthomas: I also prefer the naming convention you have suggested. I just wanted to stay consistent with the returned class name but I will change this.

I have decided to merge `getParentUri()` and `getGrandParentUri()` methods.

lumeer-core/src/main/java/io/lumeer/core/WorkspaceKeeper.java
livthomas: I have implemented the caching now.

livthomas:
I have made some changes according to your comments in the last commit.

By the way, I would love to work on the front-end all the time. However, some work still needs to be done on the back-end side otherwise we will end up with unmaintainable code base.

marvec:
I am fine with the changes, just Travis has some arguments...

marvec:
I am fine with the changes, just Travis has some arguments...

lumeer-core/src/main/java/io/lumeer/core/PermissionsChecker.java
```diff
@@ -0,0 +1,106 @@
...
+   public boolean hasRole(Resource resource, Role role) {
+      return getActualRoles(resource).contains(role);
+   }
+
+   private boolean hasUserRole(Set<Permission> userRoles, String user, Role role) {
+      return userRoles.stream()
+                      .filter(entity -> entity.getName().equals(user))
+                      .flatMap(entity -> entity.getRoles().stream())
+                      .anyMatch(actualRole -> actualRole.equals(role));
+   }
```
livthomas: Oh, these are the old methods that are no longer needed since they alternatives has already been implemented using `getActualRoles`.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoOrganizationDao.java
```diff
@@ -0,0 +1,50 @@
...
+      datastore.save(mongoOrganization);
+      return mongoOrganization;
+   }
+
+   @Override
+   public Organization getOrganizationByCode(final String organizationCode) {
+      return datastore.createQuery(MongoOrganization.class)
+                      .field(MongoOrganization.CODE).equal(organizationCode)
+                      .get();
+   }
```
kubedo8: In `MongoViewDao` is similar method implemented with checking for non null. Consider implement it with similar strategy.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoProjectDao.java
```diff
@@ -0,0 +1,50 @@
...
+      datastore.save(mongoProject);
+      return mongoProject;
+   }
+
+   @Override
+   public Project getProjectByCode(final String projectCode) {
+      return datastore.createQuery(MongoProject.class)
+                      .field(MongoProject.CODE).equal(projectCode)
+                      .get();
+   }
```
kubedo8: Also here.

lumeer-core/src/main/java/io/lumeer/core/facade/ViewFacade.java
```diff
@@ -0,0 +1,173 @@
...
+   public List<View> getAllViews() {
+      return viewDao.getAllViews().stream()
+                    .filter(view -> permissionsChecker.hasRole(view, Role.READ))
+                    .map(this::keepOnlyActualUserRoles)
+                    .collect(Collectors.toList());
+   }
+
+   private View keepOnlyActualUserRoles(View view) {
+      Set<Role> roles = permissionsChecker.getActualRoles(view);
+      Permission permission = new SimplePermission(authenticatedUser.getUserEmail(), roles);
```
livthomas: `getActualRoles` method returns all roles assigned to the currently logged in user regardless of whether they are direct or gained through the group membership. `ViewFacade` then returns a resource with only one permission record (for the authenticated user) where there are all these roles so they can be easily checked on the front-end.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoOrganizationDao.java
livthomas: Thanks, I forgot to change it here.

lumeer-core/src/main/java/io/lumeer/core/facade/ViewFacade.java
```diff
@@ -0,0 +1,173 @@
...
+
+      return keepOnlyActualUserRoles(view);
+   }
+
+   public List<View> getAllViews() {
+      return viewDao.getAllViews().stream()
+                    .filter(view -> permissionsChecker.hasRole(view, Role.READ))
+                    .map(this::keepOnlyActualUserRoles)
+                    .collect(Collectors.toList());
+   }
```
livthomas: This method was there for the demonstration how to get all entities. We will never need to get all of them so I have rewritten it using the query parameter where you can specify various properties including user and groups.

PR #97, opened 2017-08-29
Refactor documents and collections for search #LMR-307

livthomas:


lumeer-core/src/main/java/io/lumeer/core/PermissionsChecker.java
```diff
@@ -75,13 +77,20 @@ private boolean hasRole(Resource resource, Role role) {
...
 
       Set<Role> actualRoles = getActualUserRoles(resource.getPermissions().getUserPermissions(), user);
       actualRoles.addAll(getActualGroupRoles(resource.getPermissions().getGroupPermissions(), groups));
       return actualRoles;
    }
 
+   private Set<String> getUserGroups(String user, Resource resource) {
+      if (resource instanceof Organization) {
+         return Collections.emptySet();
+      }
```
kubedo8: Why is this condition here?

lumeer-core/src/main/java/io/lumeer/core/facade/AbstractFacade.java
```diff
@@ -48,11 +59,20 @@
...
       resource.getPermissions().updateUserPermissions(userPermissions.toArray(new Permission[0]));
 
       Set<Permission> groupPermissions = storedPermissions.getGroupPermissions();
       resource.getPermissions().updateGroupPermissions(groupPermissions.toArray(new Permission[0]));
    }
+
+   protected SearchQuery createSearchQuery(Pagination pagination) {
+      String user = authenticatedUser.getCurrentUsername();
+      Set<String> groups = userCache.getUser(user).getGroups();
+
```
kubedo8: Pagination is here for future use, am I right?

lumeer-core/src/main/java/io/lumeer/core/facade/OrganizationFacade.java
```diff
@@ -146,4 +140,20 @@ public void removeGroupPermission(final String organizationCode, final String gr
       organization.getPermissions().removeGroupPermission(group);
       organizationDao.updateOrganization(organization.getId(), organization);
    }
+
+   private void createOrganizationScopedRepositories(Organization organization) {
+      projectDao.setOrganization(organization);
+      projectDao.createProjectsRepository(organization);
+
```
kubedo8: Do we really need to setOrganization before creating its repository? Same with delete..

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
```diff
@@ -0,0 +1,174 @@
...
+                          .collect(Collectors.toList());
+   }
+
+   private List<Collection> getCollectionsByDocumentSearch(Query query) {
+      java.util.Collection<Collection> searchedCollections = getCollections(query.getCollectionCodes()).values();
+      List<Collection> matchedCollections = new ArrayList<>();
+
+      for (Collection collection : searchedCollections) {
+         long documentCount = dataDao.getDataCount(collection.getId(), createSearchQuery(query));
+         if (documentCount > 0) {
```
kubedo8: Why don't you use `getDocumentCount` method on Collection?

lumeer-core/src/main/java/io/lumeer/core/facade/AbstractFacade.java
```diff
@@ -19,26 +19,37 @@
...
-   protected AuthenticatedUser authenticatedUser;
+   protected UserCache userCache;
 
-   public <T extends Resource> T keepOnlyActualUserRoles(final T resource) {
+   @Inject
+   protected WorkspaceKeeper workspaceKeeper;
+
+   protected <T extends Resource> T keepOnlyActualUserRoles(final T resource) {
       Set<Role> roles = permissionsChecker.getActualRoles(resource);
       Permission permission = new SimplePermission(authenticatedUser.getUserEmail(), roles);
```
kubedo8: You often use `getCurrentUserName`

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/model/MorphiaCollection.java
```diff
@@ -0,0 +1,137 @@
...
+import org.mongodb.morphia.utils.IndexType;
+
+import java.time.LocalDateTime;
+import java.util.Collections;
+import java.util.List;
+
+@Entity
+@Indexes({
+      @Index(fields = { @Field(MorphiaCollection.CODE) }, options = @IndexOptions(unique = true)),
+      @Index(fields = { @Field(MorphiaCollection.NAME) }),
```
kubedo8: Name should be also unique, or no?

war/src/test/java/io/lumeer/core/facade/CollectionFacadeIntegrationTest.java
```diff
@@ -0,0 +1,269 @@
...
+      userDao.setOrganization(storedOrganization);
+
+      SimpleUser user = new SimpleUser(USER);
+      userDao.createUser(user);
+
+      JsonProject project = new JsonProject();
+      project.setCode(PROJECT_CODE);
+
+      JsonPermissions projectPermissions = new JsonPermissions();
+      projectPermissions.updateUserPermissions(new JsonPermission(USER, Project.ROLES.stream().map(Role::toString).collect(Collectors.toSet())));
```
kubedo8: Just a hint: You can use `Role.toStringRoles(Project.ROLES)` instead

api/src/main/java/io/lumeer/api/dto/JsonAttribute.java
```diff
@@ -0,0 +1,126 @@
...
+
+   @Override
+   public String getFullName() {
+      return fullName;
+   }
+
+   @Override
+   public Set<String> getConstraints() {
+      return constraints;
+   }
```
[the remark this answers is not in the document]
livthomas: No, because we might want to add other constraints on the facade layer. We will use `Json*` entities for abstract manipulation with the data (they are not database-specific).

lumeer-core/src/main/java/io/lumeer/core/facade/AbstractFacade.java
livthomas: No, it should have already been used. But I haven't written any tests for the pagination, that's why I haven't discovered this problem. Nice catch! Thanks.

lumeer-core/src/main/java/io/lumeer/core/facade/OrganizationFacade.java
livthomas: Yes, because projects are stored in a separate collection in the user database now. And I have used organization ID in the mongo collection name so system and user database can be the same (as we use it for testing).

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
livthomas: Which method do you mean? Document data is stored in multiple mongo collections (per our collection) and we need to search all of them within the given project.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/model/MorphiaCollection.java
livthomas: It depends. There is basically no special use of name on the back-end side which would require making it unique. However, it does make sense from the user's perspective since they only deal with the names and see codes just in the URL.

war/src/test/java/io/lumeer/core/facade/CollectionFacadeIntegrationTest.java
livthomas: Thanks. I have forgotten about that one.

api/src/main/java/io/lumeer/api/dto/JsonSuggestions.java
```diff
@@ -0,0 +1,67 @@
...
+   }
+
+   public List<JsonCollection> getCollections() {
+      return collections;
+   }
+
+   public List<JsonView> getViews() {
+      return views;
+   }
+
```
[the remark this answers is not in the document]
livthomas: Not really but I can make them so.

lumeer-core/src/main/java/io/lumeer/core/PermissionsChecker.java
livthomas: Because checking groups when dealing with organization rights does not make sense. Groups are stored per organization but you would need something above it to be able to check these access rights.

lumeer-core/src/main/java/io/lumeer/core/facade/AbstractFacade.java
livthomas: It's basically the same. The only difference is that `getUserEmail()` just gets the email from Keycloak while `getCurrentUserName()` is a shortcut for `getUser().getUsername()` and `getUser()` tries to get user record from `UserCache`. It's mainly useful for getting the groups for a given user.

PR #99, opened 2017-08-30
Create import data component #LMR-324

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/facade/ImportFacade.java
```diff
@@ -0,0 +1,187 @@
...
+      document.setCollectionId(collection.getId());
+      document.setCreatedBy(authenticatedUser.getCurrentUsername());
+      document.setCreationDate(LocalDateTime.now());
+      document.setDataVersion(DocumentFacade.INITIAL_VERSION);
+   }
+
+   private Collection createImportCollection() {
+      checkProjectWriteRole();
+      Set<String> names = collectionDao.getAllCollections().stream().map(Collection::getName).collect(Collectors.toSet());
+      Set<String> codes = collectionDao.getAllCollections().stream().map(Collection::getCode).collect(Collectors.toSet());
```
livthomas: Why do we make two database calls instead of just one? We should first load all collections and then map it to those sets.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/CollectionDao.java
```diff
@@ -43,6 +43,8 @@
 
    Collection getCollectionByCode(String code);
 
+   List<Collection> getAllCollections();
```
livthomas: We do not need this method. There is rarely a case when you really want to get all collections. We should always use `getCollections(SearchQuery query)` method with the right pagination.

lumeer-core/src/main/java/io/lumeer/core/facade/ImportFacade.java
livthomas: You can use `getAllCollectionCodes()` for this purpose. You do not really need collection names here since they are not unique. We can use the file name as a collection name for instance.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/model/MorphiaDocument.java
```diff
@@ -198,4 +201,21 @@ public String toString() {
...
    }
+
+   public org.bson.Document toBsonDocument() {
+      DataDocument dataDocument = new DataDocument(COLLECTION_ID, collectionId)
+            .append(CREATED_BY, createdBy)
+            .append(CREATION_DATE, convertLocalDateTimeToDate(creationDate))
+            .append(UPDATED_BY, updatedBy)
+            .append(UPDATE_DATE, convertLocalDateTimeToDate(updateDate))
+            .append(DATA_VERSION, dataVersion);
+      return MongoUtils.dataDocumentToDocument(dataDocument);
```
livthomas: You are in Mongo module here so you can directly create `org.bson.Document` and do not waste resources by converting it. :)

PR #101, opened 2017-09-04
Fix search results pagination

livthomas:
@justmesr, this should fix the problem you have found.


livthomas:
I'm closing this since it does not fix the issue properly and more complex solution will be required.

PR #102, opened 2017-09-04
Fix and extend organization permissions tests #LMR-312

SquattingPenguin:


war/src/test/java/io/lumeer/remote/rest/permissions/OrganizationServicePermissionsIntegrationTest.java
```diff
@@ -0,0 +1,404 @@
...
+import javax.inject.Inject;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.GenericType;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@RunWith(Arquillian.class)
+public class OrganizationServicePermissionsIntegrationTest extends IntegrationTestBase {
```
livthomas: You should inherit from the new `ServiceIntegrationTestBase` class so you don't have to create a new HTTP client in each test and can use `client` attribute from the parent class instead. It will also handle closing the client after the tests so there will be no hanging connections if a test fails.

PR #103, opened 2017-09-04
Fix and extend project permissions tests

SquattingPenguin:


war/src/test/java/io/lumeer/remote/rest/permissions/ProjectServicePermissionsIntegrationTest.java
```diff
@@ -0,0 +1,364 @@
...
+import javax.inject.Inject;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.GenericType;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@RunWith(Arquillian.class)
+public class ProjectServicePermissionsIntegrationTest extends IntegrationTestBase {
```
livthomas: The same suggestion as in the organization tests applies here as well. You can inherit from `ServiceIntegrationTestBase` and simplify tests so you don't have to initialize and close the client.

PR #108, opened 2017-11-09
Remove _id attribute before patching data

livthomas:
Without this fix, it is not possible to update the data record in the new table perspective.


livthomas:
The build fail is not related to the changes. I will investigate it separately.

PR #114, opened 2018-01-17
Store views on the server #LMR-309

livthomas:


api/src/main/java/io/lumeer/api/dto/JsonView.java
```diff
@@ -33,9 +32,11 @@
 
    public static final String QUERY = "query";
    public static final String PERSPECTIVE = "perspective";
+   public static final String CONFIG = "config";
 
    private JsonQuery query;
    private String perspective;
+   private Object config;
```
kubedo8: Why we use `Object` here?

livthomas: Because we do not need to know the structure of this field on the back-end side since it only matters for the front-end state.

kubedo8:
Would not we want to work without **Morphia**?

api/src/main/java/io/lumeer/api/dto/JsonView.java
kubedo8: Okay. Can't happen that something comes from front-end that back-end can't save?

livthomas: It's just a JSON that will go through the back-end and will be stored by MongoDB with the same structure as it was created on the front-end.

PR #118, opened 2018-01-20
Create link services #LMR-558

kubedo8:


api/src/main/java/io/lumeer/api/dto/JsonQuery.java
```diff
@@ -111,31 +158,43 @@ public boolean equals(final Object o) {
       if (this == o) {
          return true;
       }
-      if (!(o instanceof Query)) {
+      if (!(o instanceof JsonQuery)) {
```
livthomas: `Query` interface was deliberately used here in order to be able to compare the instance with `MorphiaQuery` as well. It may not be needed anymore but I still wonder why this change?

api/src/main/java/io/lumeer/api/model/LinkInstance.java
```diff
@@ -19,25 +19,94 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
-public interface LinkInstance {
+public class LinkInstance {
+
+   public static final String ID = "_ID";
```
livthomas: Why do we define `_id` field here? This class should not be mapped to MongoDB in any way.

api/src/main/java/io/lumeer/api/model/LinkInstance.java
```diff
@@ -19,25 +19,94 @@
...
+   private String id;
+   private String linkTypeId;
+   private List<String> documentIds;
+   private Map<String, Object> data;
+
+   @JsonCreator
+   public LinkInstance(@JsonProperty(LINK_TYPE_ID) final String linkTypeId,
+         @JsonProperty(DOCUMENTS_IDS) final List<String> documentIds,
+         @JsonProperty(DATA) final Map<String, Object> data) {
+      this.id = id;
```
livthomas: Where does this `id` come from?

api/src/main/java/io/lumeer/api/model/LinkType.java
```diff
@@ -19,24 +19,94 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Collections;
 import java.util.List;
 
-public interface LinkType {
+public class LinkType {
+
+   public static final String ID = "_id";
```
livthomas: The same thing here. If you want to refer to this as JSON field mapping, it should be `id` instead.

api/src/main/java/io/lumeer/api/model/LinkType.java
```diff
@@ -19,24 +19,94 @@
...
+
+   private String id;
+   private String name;
+   private List<String> collectionIds;
+   private List<JsonAttribute> attributes;
+
+   @JsonCreator
+   public LinkType(@JsonProperty(NAME) final String name,
+         @JsonProperty(COLLECTION_IDS) final List<String> collectionIds,
+         @JsonProperty(ATTRIBUTES) final List<JsonAttribute> attributes) {
```
livthomas: We would probably want to map `id` field coming from the front-end as well, or not?

lumeer-core/src/main/java/io/lumeer/core/facade/LinkInstanceFacade.java
```diff
@@ -19,27 +19,71 @@
...
+import io.lumeer.storage.api.dao.LinkInstanceDao;
+import io.lumeer.storage.api.dao.LinkTypeDao;
 
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import javax.inject.Inject;
 
 public class LinkInstanceFacade {
```
livthomas: Scope is not defined here and so `@Dependent` is used by default. We should rather explicitly define scope in order to prevent confusion.

lumeer-core/src/main/java/io/lumeer/core/facade/LinkInstanceFacade.java
```diff
@@ -19,27 +19,71 @@
...
+   private LinkTypeDao linkTypeDao;
+
+   @Inject
+   private CollectionDao collectionDao;
+
+   @Inject
+   private LinkInstanceDao linkInstanceDao;
+
+   @Inject
+   protected PermissionsChecker permissionsChecker;
```
livthomas: You don't need to inject such common things. Just extend `AbstractFacade`.

lumeer-core/src/main/java/io/lumeer/core/facade/LinkInstanceFacade.java
```diff
@@ -19,27 +19,71 @@
...
+      linkInstanceDao.deleteLinkInstance(id);
+   }
+
+   public List<LinkInstance> getLinkInstances(Query query) {
+      return linkInstanceDao.getLinkInstances(query);
    }
 
-   public List<LinkInstance> getLinkInstances(SearchQuery query) {
-      throw new UnsupportedOperationException();
+   private void checkLinkInstancePermission(Set<String> collectionIds) {
```
livthomas: Why not using `Collection<String>` here so we don't have to convert it with almost every use?

lumeer-core/src/main/java/io/lumeer/core/facade/LinkTypeFacade.java
```diff
@@ -19,27 +19,62 @@
...
 public class LinkTypeFacade {
 
+   @Inject
+   private LinkTypeDao linkTypeDao;
+
+   @Inject
+   private CollectionDao collectionDao;
+
+   @Inject
+   protected PermissionsChecker permissionsChecker;
```
livthomas: The same applies here as well. We can just extend `AbstractFacade` and get rid of injecting common stuff.

lumeer-core/src/main/java/io/lumeer/core/facade/LinkTypeFacade.java
```diff
@@ -19,27 +19,62 @@
...
+      linkTypeDao.deleteLinkType(id);
+   }
+
+   public List<LinkType> getLinkTypes(Query query) {
+      return linkTypeDao.getLinkTypes(query);
    }
 
-   public List<LinkType> getLinkTypes(SearchQuery query) {
-      throw new UnsupportedOperationException();
+   private void checkLinkTypePermission(Set<String> collectionIds) {
```
livthomas: The same applies here. `Collection<String>` would save some unnecessary conversions.

lumeer-remote/src/main/java/io/lumeer/remote/rest/LinkInstanceService.java
```diff
@@ -19,29 +19,67 @@
...
    public LinkInstance createLinkInstance(LinkInstance linkInstance) {
-      throw new UnsupportedOperationException();
+      return linkInstanceFacade.createLinkInstance(linkInstance);
    }
 
-   public LinkInstance updateLinkInstance(String id, LinkInstance linkInstance) {
-      throw new UnsupportedOperationException();
+   @PUT
+   @Path("{linkIstanceId}")
+   public LinkInstance updateLinkInstance(@PathParam("linkIstanceId") String id, LinkInstance linkInstance) {
```
livthomas: Two typos detected! :D

lumeer-remote/src/main/java/io/lumeer/remote/rest/LinkInstanceService.java
```diff
@@ -19,29 +19,67 @@
...
+      return linkInstanceFacade.updateLinkInstance(id, linkInstance);
    }
 
-   public void deleteLinkInstance(String id) {
-      throw new UnsupportedOperationException();
+   @DELETE
+   @Path("{linkIstanceId}")
+   public Response deleteLinkInstance(@PathParam("linkIstanceId") String id) {
```
livthomas: And two more...

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/LinkInstanceDao.java
```diff
@@ -20,20 +20,29 @@
...
+   void deleteLinkInstance(String id);
 
-   void deleteLinkInstance(String linkInstanceId);
+   void deleteLinkInstances(Query query);
 
-   void deleteLinkInstances(SearchQuery query);
+   LinkInstance getLinkInstance(String id);
 
-   List<LinkInstance> getLinkInstances(SearchQuery query);
+   List<LinkInstance> getLinkInstances(Query query);
```
livthomas: I think we use `SearchQuery` on DAO level.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/CollectionDao.java
```diff
@@ -42,6 +42,8 @@
 
    Collection getCollectionByCode(String code);
 
+   List<Collection> getCollectionByIds(List<String> ids);
```
livthomas: `getCollectionsByIds`

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MorphiaCollectionDao.java
```diff
@@ -92,6 +92,13 @@ public Collection getCollectionByCode(final String code) {
       return collection;
    }
 
+   @Override
+   public List<Collection> getCollectionByIds(final List<String> ids) {
```
livthomas: What about using `Collection<String>` here? I can see that we sometimes need to pass `Set` here and it doesn't really matter what collection type it is inside this method.

api/src/main/java/io/lumeer/api/dto/JsonQuery.java
kubedo8: This was auto-generated. I will fix that.

api/src/main/java/io/lumeer/api/model/LinkInstance.java
kubedo8: It is unused. I will remove it.

api/src/main/java/io/lumeer/api/model/LinkType.java
kubedo8: Yes of course

lumeer-core/src/main/java/io/lumeer/core/facade/LinkInstanceFacade.java
kubedo8: AbstractFacade contains lot of unnecessary things, but Ok, I will extend it.

lumeer-core/src/main/java/io/lumeer/core/facade/LinkTypeFacade.java
kubedo8: But I need distinct values here

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/codecs/LinkTypeCodec.java
```diff
@@ -40,6 +40,11 @@
 
 public class LinkTypeCodec implements CollectibleCodec<LinkType> {
 
+   public static final String ID = "_id";
+   public static final String NAME = "name";
+   public static final String COLLECTION_IDS = "collectionIds";
+   public static final String ATTRIBUTES = "attributes";
```
livthomas: +1 on defining it here

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MongoLinkInstanceDao.java
```diff
@@ -35,7 +36,7 @@
    public void createLinkInstanceRepository(Project project) {
       database.createCollection(databaseCollectionName(project));
 
-      MongoCollection<LinkInstance> projectCollection = databaseCollection();
+      MongoCollection<Document> projectCollection = database.getCollection(databaseCollectionName(project));
```
livthomas: Why this change?

kubedo8: It's same as in other Dao-s. It's cause test exception, because when we are creating repository on Project creation, we don't have set `Project` property here (but it's correct behavior).

PR #121, opened 2018-01-26
 Filter by document IDs in search results #LMR-676

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
```diff
@@ -60,17 +60,67 @@
...
-      if (!isCollectionSelectionOnly(query)) {
-         collections.addAll(getCollectionsByDocumentSearch(query));
+      if (query.getFulltext() != null && !query.getFulltext().isEmpty()) {
+         collections.addAll(getCollectionsByDocumentFulltext(query));
       }
 
-      collections.addAll(getCollectionsByCollectionSearch(query));
+      if(query.getDocumentIds() != null && !query.getDocumentIds().isEmpty()){
+         collections.addAll(getCollectionsByDocumentsIds(query.getDocumentIds()));
+      }
```
livthomas: I do not think it is the best idea to make separate requests for documentIDs and fulltext part of the query. The reason why we make another request based on `collectionCodes` is that collections are searched by this parameter in another database collection. But the ones searching for documents first should be merged into a single database query.

At the same time it would be worth making just document queries first and then get all collections (based on `collectionCodes` as well as retrieved documents) in a single database query. What do you think?

kubedo8: Okay, that sounds good.

kubedo8: But these two requests can't be marged. So I think It's good as it is.

kubedo8: But merging these requests will lead to unnecessary when searching by documents ids..

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
```diff
@@ -60,17 +60,80 @@
...
+         collections.addAll(getCollectionsByDocumentsSearch(query));
       }
 
-      collections.addAll(getCollectionsByCollectionSearch(query));
+      if (query.getDocumentIds() != null && !query.getDocumentIds().isEmpty()) {
+         collections.addAll(getCollectionsByDocumentsIds(query.getDocumentIds()));
+      }
+
+      if (isOnlyCollectionQuery(query) || collectionQueryIsNotEmpty(query)) {
+         collections.addAll(getCollectionsByCollectionSearch(query));
```
livthomas: Why this strange condition? I think there should either be AND operator or otherwise the first part of the condition does not make sense. Because if it is true the second one will also be true so the first part does not matter.

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
```diff
@@ -60,17 +60,80 @@
...
    public List<Document> searchDocuments(Query query) {
-      Map<String, Collection> collections = getCollections(query.getCollectionCodes());
+      Set<Document> documents = new HashSet<>();
+
+      if (query.getDocumentIds() != null && !query.getDocumentIds().isEmpty()) {
+         documents.addAll(getDocumentsByIds(query.getDocumentIds()));
+      }
+
+      if (!isOnlyDocumentsIdsQuery(query) || isEmptyQuery(query)) {
+         documents.addAll(searchDocumentsByFullText(query));
```
livthomas: Why do we want to search documents by fulltext in case of an empty query?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MorphiaCollectionDao.java
```diff
@@ -140,8 +141,18 @@ public Collection getCollectionByCode(final String code) {
       if (searchQuery.isFulltextQuery()) {
          mongoQuery.search(searchQuery.getFulltext());
       }
+      Set<ObjectId> collectionIds = searchQuery.getCollectionIds() != null ? searchQuery.getCollectionIds().stream().map(ObjectId::new).collect(Collectors.toSet()) : Collections.emptySet();
       if (searchQuery.isCollectionCodesQuery()) {
-         mongoQuery.field(MorphiaCollection.CODE).in(searchQuery.getCollectionCodes());
+         if (!collectionIds.isEmpty()) {
+            mongoQuery.or(
```
livthomas: Why this if statement? We use OR in the database query so we can always run it with both `code` and `id` conditions. Cannot we?

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
kubedo8: Because we want to return all documents on empty query (and fulltext is not applied then).

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MorphiaCollectionDao.java
kubedo8: It doesn't make sense to use or condition with only one statement (if second one is empty)

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
kubedo8: The first part of conditions will ensure, that empty query returns all collections.

livthomas: What? Do we want to run `getCollectionsByCollectionSearch(query)` if the query is empty (first condition) and also if it is not empty (second condition)? Why do we then have this `if` statement?

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
livthomas: Oh, I see. But it is not obvious at all from the code and it looks like inefficient solution at first.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MorphiaCollectionDao.java
livthomas: Ok, I have finally understood the code. But this bunch of `if` statements look really complicated and is hard to understand at first sight.

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
kubedo8: Ok, maybe names of conditions are confusing, so I will change it.

kubedo8: Is it better now?

livthomas: Oh, yes. I totally understand it now! Thanks.

PR #123, opened 2018-01-29
Make views and links search suggestions work #LMR-693

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/facade/SuggestionFacade.java
```diff
@@ -98,18 +106,50 @@ private JsonSuggestions suggestAll(String text, int limit) {
       return collections;
    }
 
+   private List<LinkType> suggestLinkTypes(String text, int limit) {
+      List<Collection> collections = collectionDao.getCollections(createSimpleSearchQuery());
```
livthomas: Why this call? Are we making link suggestions based on collection names?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/util/MongoFilters.java
```diff
@@ -85,4 +86,14 @@ public static Bson suggestionsFilter(SuggestionQuery query) {
       );
    }
 
+   public static <T> void addPaginationToSuggestionQuery(FindIterable<T> findIterable, SuggestionQuery query) {
```
livthomas: I don't think this method belongs to this class. I would suggest moving it `MongoDao` abstract class for now. And we will see how many similar methods will get there and can then extract them to a new class.

lumeer-core/src/main/java/io/lumeer/core/facade/SuggestionFacade.java
kubedo8: We need this for check which link-types can user read.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/util/MongoFilters.java
kubedo8: I thought you would not like it here 😄

lumeer-core/src/main/java/io/lumeer/core/facade/SuggestionFacade.java
livthomas: Oh, I see. I forgot about it and did not see checking of user permissions anywhere.

livthomas: Maybe it would be worth naming it `allowedCollections` or something like that.

PR #125, opened 2018-02-08
Create user and group services #LMR-686

kubedo8:


war/src/test/java/io/lumeer/remote/rest/UserServiceIntegrationTest.java
```diff
@@ -0,0 +1,171 @@
...
+   private static final String USER = AuthenticatedUser.DEFAULT_EMAIL;
+
+   private static final String USER1 = "user1@gmail.com";
+   private static final String USER2 = "user2@gmail.com";
+   private static final String USER3 = "user3@gmail.com";
+
+   private static final Set<String> GROUPS;
+
+   static {
+      GROUPS = new HashSet<>(Arrays.asList("group1", "group2", "group3"));
```
justmesr: Why is this in a `static` block?

(Was there something wrong with `private static final Set<String> GROUPS = new HashSet<>(Arrays.asList("group1", "group2", "group3"));`?)

kubedo8: I copied it from another test, but you are right there's no reason to do it in static block in this situation.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoGroupDao.java
```diff
@@ -0,0 +1,117 @@
...
+import org.bson.Document;
+import org.bson.conversions.Bson;
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.RequestScoped;
+
+@RequestScoped
+public class MongoGroupDao extends SystemScopedDao implements GroupDao {
```
livthomas: I thought we made it clear that groups will be organization-scoped. Or are there any performance benefits or other reasons why it is better to have it system-scoped? Because I cannot see any at the moment.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
```diff
@@ -0,0 +1,195 @@
...
+import org.bson.conversions.Bson;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.RequestScoped;
+
+@RequestScoped
```
livthomas: This should be `@ApplicationScoped` since it manages system-wide users and there is no point in creating a new instance for each request.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/UserDao.java
```diff
@@ -18,28 +18,23 @@
...
+   User updateUser(String organizationId, String userId, User user);
 
-   void setOrganization(Organization organization);
+   void deleteUser(String organizationId, String userId);
 
-   User createUser(User user);
+   void deleteGroupFromUsers(String organizationId, String group);
 
-   void updateUser(String id, User user);
+   Optional<User> getUserByEmail(String organizationId, String email);
```
livthomas: We don't really need to know `organizationId` in order to get a user by email, do we?

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/UserDao.java
```diff
@@ -18,28 +18,23 @@
...
 public interface UserDao {
 
-   void createUsersRepository(Organization organization);
+   User createUser(String organizationId, User user);
 
-   void deleteUsersRepository(Organization organization);
+   User updateUser(String organizationId, String userId, User user);
 
-   void setOrganization(Organization organization);
+   void deleteUser(String organizationId, String userId);
```
livthomas: Why do we need `organizationId` parameter on these three methods? It is not needed here since the DAO layer should only be responsible for storing and retrieving data. And it is system-scoped so there is no need for an organization ID here.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
```diff
@@ -0,0 +1,195 @@
...
+      database.getCollection(databaseCollectionName()).drop();
+   }
+
+   @Override
+   public User createUser(final String organizationId, final User user) {
+      MongoUser existingUser = databaseCollection().find(Filters.eq(UserCodec.EMAIL, user.getEmail())).first();
+      if (existingUser != null) {
+         return upsertUserGroupsForOrganization(organizationId, existingUser.getId(), user);
+      }
+      return createNewUser(organizationId, user);
```
livthomas: Why do we have all this "business logic" here? DAO layer should be really stupid and responsible only for storing and retrieving data. If we don't follow these rules which we set, we will eventually end up with the same spaghetti code we once had here.

You can clearly see that something is wrong just by looking at the `UserFacade` which doesn't really do anything useful (except for checking the permissions) and just calls the DAO methods.

lumeer-core/src/main/java/io/lumeer/core/cache/UserCache.java
```diff
@@ -38,25 +38,29 @@
...
-   public User getUser(String username) {
-      return userCache.computeIfAbsent(username, this::getOrCreateUser);
+   public User getUser(String email) {
+      return userCache.computeIfAbsent(email, this::getOrCreateUser);
    }
 
-   private User getOrCreateUser(String username) {
-      Optional<User> userOptional = userDao.getUserByUsername(username);
+   private User getOrCreateUser(String email) {
+      String organizationId = workspaceKeeper.getOrganization().get().getId(); // TODO how to be sure, that organization is set?
```
livthomas: Once you fix the problem with DAO methods you will see there is no need for getting the organization here.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/GroupDao.java
```diff
@@ -24,12 +24,12 @@
...
-   String createGroup(Group group);
+   Group createGroup(String organizationId, Group group);
 
-   void updateGroup(String id, Group group);
+   Group updateGroup(String id, Group group);
 
    void deleteGroup(String id);
 
-   List<Group> getAllGroups();
+   List<Group> getAllGroups(String organizationId);
```
livthomas: Once you fix the problem with the scope of this DAO, `organizationId` parameter will not be needed in any of these methods.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
```diff
@@ -0,0 +1,195 @@
...
+      if (database.getCollection(COLLECTION_NAME) == null) {
+         createUsersRepository();
+      }
+   }
+
+   public void createUsersRepository() {
+      database.createCollection(databaseCollectionName());
+
+      MongoCollection<Document> userCollection = database.getCollection(databaseCollectionName());
+      userCollection.createIndex(Indexes.ascending(UserCodec.EMAIL), new IndexOptions().unique(true));
```
livthomas: I think we would also need an index on `keycloakId` field.

lumeer-core/src/main/java/io/lumeer/core/facade/ProjectFacade.java
```diff
@@ -102,7 +102,7 @@ public Project getProject(final String projectCode) {
 
    public List<Project> getProjects() {
       User user = authenticatedUser.getCurrentUser();
-      DatabaseQuery query = DatabaseQuery.createBuilder(user.getUsername())
+      DatabaseQuery query = DatabaseQuery.createBuilder(user.getEmail())
```
livthomas: In the future, we would probably want to use user ID for this. But it is out of the scope of this PR.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/model/MongoGroup.java
```diff
@@ -0,0 +1,50 @@
+package io.lumeer.storage.mongodb.model;
+
+import io.lumeer.api.model.Group;
+
+public class MongoGroup {
+
+   private String id;
+   private String name;
+   private String organizationId;
```
livthomas: Once the problem with the scope of `MongoGroupDao` is fixed, `organizationId` will no longer be needed here.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoGroupDao.java
kubedo8: Each organization will contain only few groups. Do we really want to have so many small collections?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
kubedo8: Okay

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoGroupDao.java
livthomas: Good point. But how do you then want to deal with unique group names just within a single organization if all groups are stored in a single collection?

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/UserDao.java
kubedo8: Look at how users are stored in DB. How can I know, which embedded document from array of groups should I delete?

livthomas: Yes, but you don't need it on this level. I mentioned it in the other comment as well. This business logic needs to be in the facade not in DAO.

Deleting user on DAO level means deleting it from the database. (Which is not really something we need at the moment.)

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
kubedo8: It's more complicated than you think. Creating users is not always creating, because when user already exist in db, we just add information that he is in some organization (by its id). So first time it is real create in db, and next time we do update. So who should be responsible for knowing structure of how users are stored in db?

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/UserDao.java
kubedo8: Maybe I should rename method name to `deleteUserFromOrganization`, but without this parameter `dao` layer doesn't know how to build correct request do db.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoGroupDao.java
kubedo8: It's simple: just add compound unique index on `organizationId` and `groupName`. It's also tested. (look at `MongoGroupDaoTest#testCreateGroupAnotherOrganization` method)

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
kubedo8: Oh good catch.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/UserDao.java
kubedo8: We need to know `organizationId` to obtain correct groups for user.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/UserDao.java
livthomas: And what about looking at this from another perspective? DAO is only responsible for simple data operations. If you want to have a method like `deleteUserFromOrganization`, it should be on the facade level and calling `updateUser` or something similar.

Or it can be like you suggest but make sure there absolutely no business logic involved in the DAO layer. All of it needs to be in the facade and DAO methods should be called there only where you would put a single database operation if you used MongoDB driver directly in the facades.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/dao/UserDao.java
livthomas: Oh, I see. I'm just wondering if we cannot do this kind of logic in the facade. It's what facades are there for. I know we would read more data from the database but how many organizations will users be in? One, two?

I think the same principle is used with permissions which are filtered on the facade layer.

The other think is that we cache the user so what happens if he switch the organization within the same session? I'm afraid the necessary data will not be in the cache since they were filtered by DAO layer.

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoGroupDao.java
livthomas: And what about our idea of storing organization data in different MongoDB databases and having only really necessary collections in the system database?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
livthomas: Creating user on the DAO layer should be only about creating the database record. Creating user on the facade and service layer should be a more complex operation (what front-end expects and what you have implemented in the DAO).

The only thing that needs to be done is moving the content this method to the facade and replace database calls by DAO method calls. If the user exists, you will update it. If not, you will creat it. Is that clear?

I'm just trying to avoid the situation we had here half a year ago. If we don't follow these simple rules, after a few such changes we will be back there and every change will be a pain again.

livthomas: What I'm trying to tell you is that there does not always needs to be 1-to-1 mapping between the DAO and facade methods. They can sometimes do something slightly different. Otherwise, we would not need these distinct layers.

lumeer-core/src/main/java/io/lumeer/core/facade/GroupFacade.java
```diff
@@ -41,33 +43,46 @@
...
    }
 
-   public void deleteGroup(String organizationId, String groupId){
-      checkPermissions(organizationId);
+   public void deleteGroup( String groupId){
+      checkPermissions();
 
-      userDao.deleteGroupFromUsers(organizationId, groupId);
       groupDao.deleteGroup(groupId);
+      deleteInfoAboutGroup(groupId);
```
livthomas: Isn't deleting the group in the line above this one going to remove all information about the group? I simply don't know what this method does unless I look into its implementation. And that's exactly what we have discussed with @marvec. If we name methods like this, we will definitely need comments in our code. But we can do better and avoid them. :)

I would suggest renaming it to `removeGroupFromUsers()` but feel free to find a better name if you wish.

lumeer-core/src/main/java/io/lumeer/core/AuthenticatedUser.java
```diff
@@ -42,11 +46,24 @@
...
 
+   @Inject
+   private WorkspaceKeeper workspaceKeeper;
+
    public User getCurrentUser() {
       String userEmail = getUserEmail();
       return userCache.getUser(userEmail);
    }
 
+   public Set<String> getCurrentUserGroups(){
```
livthomas: +1 on extracting this into a separate method

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserDao.java
```diff
@@ -74,72 +73,39 @@ public void deleteUsersRepository() {
...
-      Bson filter = Filters.and(idFilter(userId), Filters.ne(MongoUtils.concatParams(UserCodec.ALL_GROUPS, UserCodec.ORGANIZATION_ID), organizationId));
-      FindOneAndUpdateOptions options = new FindOneAndUpdateOptions().returnDocument(ReturnDocument.AFTER);
+   public User updateUser(final String id, final User user) {
+      FindOneAndReplaceOptions options = new FindOneAndReplaceOptions().returnDocument(ReturnDocument.AFTER).upsert(true);
       try {
-         MongoUser returnedUser = databaseCollection().findOneAndUpdate(filter, update, options);
-         if (returnedUser == null) {
-            return setUserGroupsForOrganization(organizationId, userId, user);
+         User returnedGroup = databaseCollection().findOneAndReplace(idFilter(id), user, options);
+         if (returnedGroup == null) {
```
livthomas: Just a detail but it's `returnedUser` I suppose. :)

PR #128, opened 2018-02-19
Send collection info on import #LMR-740

kubedo8:


kubedo8:
There's no special reason for having `format` inside the class. So I will change it in a moment.

kubedo8:
Now it should be okay.

PR #135, opened 2018-03-16
Save attributes info on import collection & apply pagination on search documents #LMR-765 #LMR-766

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/facade/ImportFacade.java
```diff
@@ -63,43 +68,17 @@
    private DataDao dataDao;
 
    public Collection importDocuments(String format, ImportedCollection importedCollection) {
-      List<Document> documents;
+      Collection collection = createImportCollection(importedCollection.getCollection());
```
livthomas: Isn't this method the same as `CollectionFacade.createCollection`? If yes, I think we can make an exception and call the other facade here in order to eliminate code duplication.

kubedo8: You told me, that facade should never call facade... Why now it's OK?!

PR #147, opened 2018-04-07
Using id instead of email/name in Users/Groups #LMR-767

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/facade/UserFacade.java
```diff
@@ -50,10 +50,14 @@ public User createUser(String organizationId, User user) {
       User storedUser = userDao.getUserByEmail(user.getEmail());
 
       if (storedUser == null) {
-         return userDao.createUser(user);
+         User createdUser = userDao.createUser(user);
+         addDefaultPermissions(organizationId, createdUser.getId());
```
livthomas: Should not permissions be added before you store the entity into the database in order not to make two database requests?

lumeer-core/src/main/java/io/lumeer/core/facade/UserFacade.java
```diff
@@ -117,4 +121,8 @@ private void checkOrganizationInUser(String organizationId, User user) {
       }
    }
 
+   private void addDefaultPermissions(String organizationId, String userId) {
+      // TODO
```
livthomas: What is the implementation of this method going to look like?

lumeer-remote/src/main/java/io/lumeer/remote/rest/GlobalService.java
```diff
@@ -0,0 +1,46 @@
...
+@Path("/global")
+public class GlobalService extends AbstractService {
+
+   @Inject
+   private AuthenticatedUser authenticatedUser;
+
+   @GET
+   @Path("currentUser")
+   public User getCurrentUser() {
+      return authenticatedUser.getCurrentUser();
```
livthomas: Why do we need a new `GlobalService`. Would not it be more better to have this method in `UserService`?

PR #161, opened 2018-05-03
Generate unique immutable attribute IDs #LMR-789

kubedo8:


api/src/main/java/io/lumeer/api/dto/JsonCollection.java
```diff
@@ -43,6 +44,9 @@
    private Integer documentsCount;
    private LocalDateTime lastTimeUsed;
 
+   private Integer lastAttributeNum;
+   private String attributePrefix;
```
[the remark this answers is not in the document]
kubedo8: Each collection has different prefix on attributes ids. This will ensure that they are unique across the db.

lumeer-core/src/main/java/io/lumeer/core/facade/DocumentFacade.java
```diff
@@ -169,6 +171,8 @@ private void updateCollectionMetadata(Collection collection, Set<String> attribu
          }
 
       });
+      System.out.println("LMRDF adding " + attributesToInc.toString());
```
livthomas: These debug outputs are probably not needed anymore.

kubedo8: Oh yes, I will delete it.

PR #167, opened 2018-05-17
Add suport for setting defaultAttribute #LMR-790

kubedo8:


api/src/main/java/io/lumeer/api/dto/JsonCollection.java
```diff
@@ -102,8 +104,8 @@ public void updateAttribute(final String attributeId, final Attribute attribute)
    }
 
    @Override
-   public void deleteAttribute(final String attributeName) {
-      attributes.removeIf(attribute -> AttributeUtil.isEqualOrChild(attribute, attributeName));
+   public void deleteAttribute(final String attributeId) {
+      attributes.removeIf(attribute -> attribute.getId().equals(attributeId));
```
livthomas: This change breaks deleting child attributes.

kubedo8: Really? How do we know which are child attributes?

livthomas: Those which names start with the name of the deleted attribute + `.`. It was easier before. Now we first need to get the deleted attribute and then remove its children.

kubedo8: Oh sorry, I misunderstood how we will use `name` field. I will definetely revert this change.

api/src/main/java/io/lumeer/api/model/Collection.java
```diff
@@ -46,7 +46,7 @@ default ResourceType getType() {
 
    void updateAttribute(String attributeId, Attribute attribute);
 
-   void deleteAttribute(String attributeId);
+   void deleteAttribute(String attributeName);
```
livthomas: Don't we pass `attributeId` to this method in `CollectionFacade`. The method interface can stay as you designed it, you just need to change the implementation. Find the attribute to be removed first in order to get the name and then remove it together with its children.

PR #187, opened 2018-07-13
Expose HTTP Location header when using CORS

livthomas:


livthomas:
I have found a better solution of this problem.

PR #196, opened 2018-08-31
Store multiple social logins for single user #LMR-1015

livthomas:


livthomas:
Do not merge yet! There is still one problem to be resolved.

PR #199, opened 2018-09-17
Make it impossible to delete user's default organization #LMR-1131

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/facade/ProjectFacade.java
```diff
@@ -93,6 +93,7 @@ public Project updateProject(final String projectCode, final Project project) {
       permissionsChecker.checkRole(storedProject, Role.MANAGE);
 
       keepStoredPermissions(project, storedProject.getPermissions());
+      keepUnmodifiableFields(project, storedProject);
```
livthomas: In the future, we should consider adding modified fields coming from the method parameter to the stored entity instead of updating the incoming entity by fields that cannot be changed. Of course, there should be some filter on incoming fields so not everything can be changed.

PR #215, opened 2018-10-20
Fix empty view id

marvec:
Fix user group handling for empty user id, self heal missing view author id for the views to be shareable, remove unused ViewDao method

### DIFF
--- a/lumeer-core/src/main/java/io/lumeer/core/facade/ViewFacade.java
+++ b/lumeer-core/src/main/java/io/lumeer/core/facade/ViewFacade.java
@@ -101,6 +101,13 @@ public class ViewFacade extends AbstractFacade {
       final View view = viewDao.getViewByCode(code);
       permissionsChecker.checkRole(view, Role.READ);
 
+      if (view.getAuthorId() == null || "".equals(view.getAuthorId())) {
+         if (permissionsChecker.hasRole(view, Role.MANAGE)) {
+            view.setAuthorId(authenticatedUser.getCurrentUserId());
+            viewDao.updateView(view.getId(), view);
+         }
+      }
+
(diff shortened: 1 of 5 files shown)


lumeer-core/src/main/java/io/lumeer/core/facade/ViewFacade.java
```diff
@@ -101,6 +101,13 @@ public View getViewByCode(final String code) {
       final View view = viewDao.getViewByCode(code);
       permissionsChecker.checkRole(view, Role.READ);
 
+      if (view.getAuthorId() == null || "".equals(view.getAuthorId())) {
+         if (permissionsChecker.hasRole(view, Role.MANAGE)) {
+            view.setAuthorId(authenticatedUser.getCurrentUserId());
+            viewDao.updateView(view.getId(), view);
+         }
+      }
```
kubedo8: This method (getViewByCode) is never called in frontend so this code doesn't heal anything.

marvec: It is called in ViewService in getView(). This is not used in frontend?

kubedo8: No

marvec: So what method is used to actually read the complete view?

marvec: And how about now?

kubedo8: It's better now :)

PR #226, opened 2018-11-25
 New concept of query #LMR-1258

kubedo8:


kubedo8:
DO NOT MERGE this PR until frontend part is implemented!!!

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
```diff
@@ -61,306 +65,258 @@
...
 
-      if (query.getDocumentIds() != null && !query.getDocumentIds().isEmpty()) {
-         collections.addAll(getCollectionsByDocumentsIds(query.getDocumentIds()));
-      }
-
-      if (collectionQueryIsNotEmpty(query) || isEmptyQuery(query)) {
-         collections.addAll(getCollectionsByCollectionSearch(query, view));
+      final View view = permissionsChecker.getActiveView();
+      if (view != null) {
+         userCollections.addAll(viewFacade.getViewsCollections(Collections.singletonList(view), true));
```
livthomas: Why cannot this be done by using directly `ViewDao`? We should try to minimize the number of dependencies between facades.

lumeer-remote/src/main/java/io/lumeer/remote/rest/OrganizationService.java
```diff
@@ -186,4 +190,11 @@ public CompanyContact setCompanyContact(@PathParam("organizationCode") final Str
       return companyContactFacade.setCompanyContact(organizationFacade.getOrganization(organizationCode), companyContact);
    }
 
+   @GET
+   @Path("convert/views")
+   public Response convertViews() {
+      viewConvertFacade.convertViews();
+      return Response.ok().build();
```
livthomas: When is this endpoint called?

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/query/SearchQuery.java
```diff
@@ -18,73 +18,26 @@
...
-   }
-
-   public boolean isFiltersQuery() {
-      return filters != null && !filters.isEmpty();
-   }
-
-   public boolean isBasicQuery() {
-      return !isFulltextQuery()  && !isLinkTypeIdsQuery() && !isDocumentIdsQuery() && !isCollectionIdsQuery();
+   public List<SearchQueryStem> getStems() {
+      return stems;
```
livthomas: Since this class is annotated as `@Immutable`, we should return immutable list here:
```java
return Collections.unmodifiableList(stems);
```

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/query/SearchQueryStem.java
```diff
@@ -0,0 +1,158 @@
...
+import io.lumeer.api.model.QueryStem;
+import io.lumeer.storage.api.filter.AttributeFilter;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class SearchQueryStem {
```
livthomas: If we want to have the whole `Query` immutable, we need to make stems immutable as well by returning unmodifiable collections (and annotate it just for record). If we need to modify the query directly somewhere and it will take some time to rewrite it, we can skip this whole immutability thing altogether and remove all traces of it.

lumeer-remote/src/main/java/io/lumeer/remote/rest/OrganizationService.java
kubedo8: This is temporary endpoint for converting old query structure to new one.

kubedo8: This will be deleted together with `ViewConvertFacade` after we convert Devel and Production database.

lumeer-core/src/main/java/io/lumeer/core/facade/SearchFacade.java
kubedo8: Because this method use multiple daos to achieve task.

PR #231, opened 2018-12-04
Implement notifications #LMR-1291

marvec:


api/src/main/java/io/lumeer/api/model/UserNotification.java
```diff
@@ -0,0 +1,179 @@
...
+   private ZonedDateTime firstReadAt;
+
+   private NotificationType type;
+
+   private DataDocument data;
+
+   public UserNotification() {
+   }
+
+   public UserNotification(final String userId, final ZonedDateTime createdAt, final boolean read, final ZonedDateTime firstReadAt, final NotificationType type, final DataDocument data) {
```
kubedo8: I think you should implement constructor with @JsonCreator annotation and @JsonProperty annonations for properties in order to work properly through Rest

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MongoCollectionDao.java
```diff
@@ -91,12 +91,13 @@ public Collection updateCollection(final String id, final Collection collection)
       FindOneAndReplaceOptions options = new FindOneAndReplaceOptions().returnDocument(ReturnDocument.AFTER);
 
       try {
-         Collection updatedCollection = databaseCollection().findOneAndReplace(idFilter(id), collection, options);
+         final Collection storedCollection = databaseCollection().find(idFilter(id)).first();
```
kubedo8: Why you don't use collection from method parameter?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserNotificationDao.java
```diff
@@ -0,0 +1,107 @@
...
+      createUserNotificationsRepository();
+   }
+
+   @Override
+   public List<UserNotification> getRecentNotifications(final String userId) {
+      Bson idFilter = Filters.eq(UserNotificationCodec.USER_ID, userId);
+
+      return databaseCollection()
+            .find(idFilter)
+            .limit(500)
```
kubedo8: 500? Are you sure?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserNotificationDao.java
```diff
@@ -0,0 +1,107 @@
...
+
+   @Override
+   public UserNotification updateNotification(final UserNotification notification) {
+      FindOneAndReplaceOptions options = new FindOneAndReplaceOptions().returnDocument(ReturnDocument.AFTER).upsert(true);
+      try {
+         UserNotification returnedNotification = databaseCollection().findOneAndReplace(idFilter(notification.getId()), notification, options);
+         if (returnedNotification == null) {
+            throw new StorageException("Notification '" + notification.getId() + "' has not been updated.");
+         }
+         // TODO send push notification
```
kubedo8: Why we should send push notification there?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserNotificationDao.java
```diff
@@ -0,0 +1,107 @@
...
+
+   @Override
+   public List<UserNotification> createNotificationsBatch(final List<UserNotification> notifications) {
+      databaseCollection().insertMany(notifications);
+      return notifications;
+   }
+
+   @Override
+   public void createUserNotificationsRepository() {
+      if (database.getCollection(COLLECTION_NAME) == null) {
```
kubedo8: This is not correct. You see there warning `Condition .... is always false`.
Why you do not look at existing solution in other SystemScopedDao? For example: `MongoOrganizationDao`

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MongoCollectionDao.java
marvec: Because it is already updated. There is no other way to get to the old value unless I want to make a copy on all places where the method is called :-(

api/src/main/java/io/lumeer/api/model/UserNotification.java
marvec: I forgot!

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserNotificationDao.java
marvec: Let's be generous ;-) What it is, like 10kB?

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserNotificationDao.java
marvec: Correct! I added push notifications handling already and not at this place!

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/system/MongoUserNotificationDao.java
marvec: Why I do not look? Well, the answer is simple, I did not know what I do not know. Then it is hard to look anywhere...

marvec: But! I copied this from some other Dao! I need to find it!

marvec: It was on 3 other places. I fixed it! Thanks for spotting this!

lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MongoCollectionDao.java
kubedo8: But why it works in other places (ProjectDao, OrganizationDao, ViewDao) without this?

marvec: It does not, the original object is simply passed as the method parameter.

marvec: In `MongoProjectDao` you can find `public Project updateProject(final String projectId, final Project project, final Project originalProject)`

kubedo8: So you can add similar method for `CollectionDao`.

marvec: I really tried at first. Unfortunately, in this case it is not that simple. Try to lookup for usages and see where the original Collection is available and where not. The CollectionDao is used on quite a lot of places and we would need to pass the original collection through several method calls typicaly. Do you think it is worth it? Would that be a better approach? Definitely more DB efficient. WDYT?

lumeer-core/src/main/java/io/lumeer/core/facade/UserNotificationFacade.java
```diff
@@ -0,0 +1,144 @@
...
+import javax.enterprise.event.Observes;
+import javax.inject.Inject;
+
+/**
+ * @author <a href="mailto:marvenec@gmail.com">Martin Večeřa</a>
+ */
+@RequestScoped
+public class UserNotificationFacade extends AbstractFacade {
+
+   private Logger log = Logger.getLogger(UserNotificationFacade.class.getName());
```
livthomas: Wouldn't it be better to declare this logger as static field? According to [SLF4J FAQ](https://www.slf4j.org/faq.html#declared_static) there is less CPU and memory overhead. Because a new instance of this facade is created with every request.

PR #242, opened 2018-12-13
Make it possible to remove user notifications #LMR-1324

marvec:


lumeer-remote/src/main/java/io/lumeer/remote/rest/NotificationService.java
```diff
@@ -54,4 +56,14 @@
    public UserNotification updateNotification(@PathParam("notificationId") String notificationId, UserNotification notification) {
       return userNotificationFacade.updateNotification(notificationId, notification);
    }
+
+   @DELETE
+   @Path("{notificationId}")
+   public Response deleteCollection(@PathParam("notificationId") String notificationId) {
```
livthomas: `deleteNotification`?

PR #266, opened 2019-02-11
Fix Blockly glitches corresponding to UI usage #LMR-1422

marvec:


lumeer-core/src/main/java/io/lumeer/core/task/executor/BlocklyRuleTaskExecutor.java
```diff
@@ -68,7 +72,12 @@ private LumeerBridge(final RuleTask ruleTask) {
       }
 
       public void setDocumentAttribute(DocumentBridge d, String attrId, Value value) {
-         changes.add(new DocumentChange(d.document, attrId, convertValue(value)));
+         try {
+            changes.add(new DocumentChange(d.document, attrId, convertValue(value)));
+         } catch (Exception e) {
+            e.printStackTrace();
+            throw e;
```
livthomas: If the exception is propagated further up the call stack, isn't it printed somewhere else anyway?

marvec: Although the short answer is unfortunately NO, the lines were there only for debugging purposes and I forgot to remove it in all methods.

PR #276, opened 2019-02-24
Add attribute to dismiss wizard to user #LMR-1485

marvec:


api/src/main/java/io/lumeer/api/model/User.java
```diff
@@ -212,13 +208,42 @@ public String toString() {
...
+            Objects.equals(agreement, user.agreement) &&
+            Objects.equals(agreementDate, user.agreementDate) &&
+            Objects.equals(newsletter, user.newsletter) &&
+            Objects.equals(wizardDismissed, user.wizardDismissed) &&
+            Objects.equals(wishes, user.wishes);
+   }
+
+   @Override
+   public int hashCode() {
+      return Objects.hash(id, name, email, authIds, groups, defaultWorkspace, agreement, agreementDate, newsletter, wizardDismissed, wishes);
```
livthomas: Should not just `id` be used in `hashCode` and `equals` methods in such entities?

marvec: Don't know, should it? What if it is not yet initialized and id is null?

livthomas: It should be some kind of a unique identifier. But you are right, `id` might be null. If `email` is not always provided either, then probably `authIds` is the best identifier. What do you think?

PR #283, opened 2019-03-02
Add link support to function tasks #LMR-1481

marvec:


lumeer-core/src/main/java/io/lumeer/core/task/executor/FunctionTaskExecutor.java
```diff
@@ -65,8 +80,12 @@ private void checkErrorErasure() {
          task.getFunction().setErrorReport("");
          task.getFunction().setTimestamp(0L);
 
-         task.getDaoContextSnapshot().getCollectionDao().updateCollection(collection.getId(), collection, null);
-         // we won't send push notifications as this is not important, it gets updated eventually
+         if (linkType == null) {
```
kubedo8: This is not very safe.
```
if (collection != null) {
....
} else if (linkType != null {
... 
}
```

lumeer-core/src/main/java/io/lumeer/core/task/executor/FunctionTaskExecutor.java
```diff
@@ -82,9 +101,13 @@ private void writeTaskError(final Exception e, final Exception cause) {
       task.getFunction().setErrorReport(sb.toString());
       task.getFunction().setTimestamp(System.currentTimeMillis());
 
-      task.getDaoContextSnapshot().getCollectionDao().updateCollection(collection.getId(), collection, null);
-
-      task.sendPushNotifications(collection);
+      if (linkType == null) {
```
kubedo8: Same here.

PR #290, opened 2019-03-12
Add computation of function dependency order #LMR-1538

marvec:


lumeer-core/src/main/java/io/lumeer/core/util/FunctionOrder.java
```diff
@@ -0,0 +1,77 @@
...
+            result.add(param);
+         }
+      });
+
+      final Deque<FunctionParameter> filteredResult = new LinkedList<>();
+      result.stream()
+            .filter(param -> dependencies.containsKey(param))
+            .forEachOrdered(param -> filteredResult.add(param));
+
+      return filteredResult;
```
kubedo8: What about this:
```
return result.removeIf(param -> !dependencies.containsKey(param));
```

PR #300, opened 2019-03-20
Execute Rules after Functions and execute Functions on document delete #LMR-1557

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/task/FunctionTask.java
```diff
@@ -90,8 +89,13 @@ public LinkType getLinkType() {
       return linkInstances;
    }
 
-   public List<FunctionTask> getParents() {
-      return parents;
+   @Override
```
marvec: I would move the parent to the Task interface and these methods to the AbstractContextualTask.

war/src/test/java/io/lumeer/core/facade/CollectionFacadeIT.java
```diff
@@ -522,10 +521,4 @@ public void testGetAllCollectionsProjectManager() {
       assertThat(collectionFacade.getCollections()).isEmpty();
    }
 
-   @Test
-   @Ignore("This test actually does not test anything, it just demonstrates the usage of tasks.")
-   public void testTaskExecutor() throws InterruptedException {
-      taskExecutor.submitTask(contextualTaskFactory.getInstance(ListCollectionsIn10SecondsTask.class));
-      Thread.sleep(15_000);
-   }
```
marvec: Please keep it there for debugging and testing purposes during development.

war/src/test/java/io/lumeer/core/task/ListCollectionsIn10SecondsTask.java
```diff
@@ -1,39 +0,0 @@
-/*
```
marvec: Keep this file please.

PR #306, opened 2019-04-06
Apply link filters & apply constraints in filters #LMR-1478

kubedo8:


api/src/main/java/io/lumeer/api/model/LinkAttributeFilter.java
```diff
@@ -1,71 +1,41 @@
...
-      final LinkAttributeFilter that = (LinkAttributeFilter) o;
-      return Objects.equals(getLinkTypeId(), that.getLinkTypeId()) &&
-            Objects.equals(getAttributeId(), that.getAttributeId()) &&
-            Objects.equals(getOperator(), that.getOperator()) &&
-            Objects.equals(getValue(), that.getValue());
-   }
-
-   @Override
-   public int hashCode() {
-      return Objects.hash(getLinkTypeId(), getAttributeId(), getOperator(), getValue());
```
livthomas: Should not both `LinkAttributeFilter` and `CollectionAttributeFilter` override `equals`, `hashCode`, and `toString` methods. It is just a detail but it might be useful while debugging some related code in the future.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/filter/CollectionAttributeFilter.java
```diff
@@ -0,0 +1,36 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package io.lumeer.storage.api.filter;
+
+import io.lumeer.api.model.ConditionType;
+
+public class CollectionAttributeFilter extends AttributeFilter {
```
livthomas: I'm a little bit confused here. I can see that we have two classes called `CollectionAttributeFilter` and `LinkAttributeFilter`.

api/src/main/java/io/lumeer/api/model/LinkAttributeFilter.java
kubedo8: Ok, I'll add it.

lumeer-storage/lumeer-storage-api/src/main/java/io/lumeer/storage/api/filter/CollectionAttributeFilter.java
kubedo8: One is for `search` filter purpose and the other is for rest & db purpose. I will rename them to make it clear what it does.

PR #333, opened 2019-06-19
Atomic attribute number assignment #LMR-1819

marvec:


lumeer-storage/lumeer-storage-mongodb/src/main/java/io/lumeer/storage/mongodb/dao/project/MongoCollectionDao.java
```diff
@@ -70,6 +71,7 @@
 public class MongoCollectionDao extends ProjectScopedDao implements CollectionDao {
 
    private static final String PREFIX = "collections_p-";
+   private static final Random rnd = new Random();
```
kubedo8: I think it was for debugging here :)

marvec: nope, it used to be for a function, but that functionality is removed ;-)

war/src/test/java/io/lumeer/core/facade/CollectionFacadeIT.java
```diff
@@ -290,6 +289,42 @@ public void testUpdateCollectionAttributeAdd() {
...
+            t.join();
+         } catch (InterruptedException e) {
+            e.printStackTrace();
+         }
+      });
+
+
+      var updatedCollection = collectionDao.getCollectionByCode(CODE);
+      assertThat(updatedCollection).isNotNull();
+      assertThat(updatedCollection.getAttributes()).hasSize(4*20);
```
livthomas: Should not we test if all their IDs are unique? Because now we only check if all attributes get created.

marvec: True, I wanted to have it stored in a set, not a list...

PR #334, opened 2019-06-20
Coordinates constraint #LMR-1528

marvec:


lumeer-core/src/main/java/io/lumeer/core/util/coordinates/CoordinatesParser.java
```diff
@@ -0,0 +1,251 @@
...
+      }
+      return Optional.empty();
+   }
+
+   private static Optional<LatLng> validateAndRound(double lat, double lon) {
+      // round to 6 decimals
+      final double latOrig = lat;
+      final double lngOrig = lon;
+      lat = roundTo6decimals(lat);
+      lon = roundTo6decimals(lon);
```
livthomas: Why are we rounding it here to 6 digits? Should not user be able to specify the precision?

marvec: The sixth decimal place is worth up to 0.11 m: you can use this for laying out structures in detail, for designing landscapes, building roads. It should be more than good enough for tracking movements of glaciers and rivers. This can be achieved by taking painstaking measures with GPS, such as differentially corrected GPS.
The seventh decimal place is worth up to 11 mm: this is good for much surveying and is near the limit of what GPS-based techniques can achieve.
The eighth decimal place is worth up to 1.1 mm: this is good for charting motions of tectonic plates and movements of volcanoes. Permanent, corrected, constantly-running GPS base stations might be able to achieve this level of accuracy.

PR #341, opened 2019-06-28
Fix templates loading #LMR-1872

marvec:


lumeer-core/src/main/java/io/lumeer/core/client/mapquest/MapQuestClient.java
```diff
@@ -22,6 +22,8 @@
 import io.lumeer.api.model.geocoding.GeoCodingResult;
 import io.lumeer.core.facade.configuration.DefaultConfigurationProducer;
 
+import org.assertj.core.util.Lists;
```
livthomas: I'm afraid this is a wrong import. The package is not available at runtime.

lumeer-core/src/main/java/io/lumeer/core/client/mapquest/MapQuestClient.java
```diff
@@ -56,6 +58,10 @@ public void init() {
    }
 
    public List<GeoCodingResult> batchGeoCode(Set<String> locations) {
+      if (mapQuestKey == null || "".equals(mapQuestKey)) {
+         return Lists.emptyList();
```
livthomas: We should use `Collections.emptyList()` instead.

PR #357, opened 2019-08-06
Supply template fix and link duplication #LMR-1953

marvec:


lumeer-core/src/main/java/io/lumeer/core/facade/LinkInstanceFacade.java
```diff
@@ -257,6 +257,29 @@ public LinkInstance getLinkInstance(String linkTypeId, String linkInstanceId) {
...
+
+         if (newLink != null) {
+            final DataDocument originalData = linkDataDao.getData(link.getLinkTypeId(), id);
+            final DataDocument newData = linkDataDao.createData(link.getLinkTypeId(), newLink.getId(), originalData);
+            constraintManager.decodeDataTypes(linkType, newData);
+            newLink.setData(newData);
+
+            result.add(newLink);
+         }
+      });
```
kubedo8: Imagine situation, when document has i.e. 1000 documents. This code call database 4000 times :( 
I know we don't have solution for problems like this, but I'm afraid we'll have big performance problems soon.

marvec: There is no other possibility, I cannot load all the data into memory. I could load all link instances, but definitely not data... So WDYT?

marvec: Is it better now?

PR #358, opened 2019-08-06
Sort query hints by string similarity #LMR-1918

kubedo8:


lumeer-core/src/main/java/io/lumeer/core/facade/SuggestionFacade.java
```diff
@@ -112,8 +113,11 @@ private boolean shouldSuggestViews(SuggestionQuery suggestionQuery) {
    }
 
    private List<View> suggestViews(SuggestionQuery suggestionQuery, int limit) {
-      SearchSuggestionQuery searchSuggestionQuery = createSuggestionQuery(suggestionQuery, limit);
-      return viewDao.getViews(searchSuggestionQuery, isManager());
+      var distance = LevenshteinDistance.getDefaultInstance();
```
livthomas: I don't know how heavy this object is and how it works internally but I just wonder if it cannot be stored as a static attribute so it doesn't have to be created every time this method is called.

kubedo8: `getDefaultInstance()` returns static instance under the hood.

PR #598, opened 2020-05-22
Add email verification status to user object

marvec:


lumeer-core/src/main/java/io/lumeer/core/auth/Auth0Filter.java
```diff
@@ -314,13 +315,16 @@ private User getUserInfo(final String accessToken) throws Auth0Exception {
       final AuthAPI auth0 = new AuthAPI(domain, clientId, clientSecret);
       final Request<UserInfo> info = auth0.userInfo(accessToken);
       final Map<String, Object> values = info.execute().getValues();
+      System.out.println(values);
```
kubedo8, line 318: Println :)

PR #621, opened 2020-07-06
Skip checking function and rules limits when creating a project from template

marvec:


lumeer-core/src/main/java/io/lumeer/core/facade/TemplateFacade.java
```diff
@@ -74,16 +78,34 @@ public String getTemplateOrganizationId(final Language language) {
...
       final TemplateParser templateParser = new TemplateParser(templateType, language);
 
-      installTemplate(project, templateParser, createTemplateMetadata(new Date()));
+      installTemplate(project, templateParser, createTemplateMetadata(new Date()), true);
    }
 
-   public void installTemplate(final Project project, final ProjectContent projectContent, final Date relativeDate) {
+   public void installTemplate(final Project project, final String organizationId, final ProjectContent projectContent, final Date relativeDate) {
       final TemplateParser templateParser = new TemplateParser(projectContent);
+      final boolean originalLumeerTemplate = getAllTemplateOrganizationIds().contains(organizationId);
```
kubedo8, line 106: I think this code is not needed, because in `CopyFacade` we can easily distinguish between Template creation and Copying project

PR #657, opened 2020-12-06
Add delayed actions

marvec:
delayed actions, delayed action processor, user notification settings, and collection purpose


api/src/main/java/io/lumeer/api/model/UserNotification.java
```diff
@@ -185,11 +181,25 @@ public int hashCode() {
...
+      String DOCUMENT_CURSOR = "documentCursor";
+   }
+
+   public interface TaskUpdated extends TaskAssigned {}
+
+   public interface DueDateSoon extends TaskAssigned {}
+
+   public interface PastDueDate extends TaskAssigned {}
+
+   public interface TaskUnassigned extends TaskAssigned {}
```
kubedo8, line 204: `TaskAssigned` as supertype is quite misleading

lumeer-core/src/main/java/io/lumeer/core/facade/UserNotificationFacade.java
```diff
@@ -126,6 +137,9 @@ public UserNotification updateNotification(final String notificationId, final Us
          data.append(UserNotification.CollectionShared.COLLECTION_NAME, resource.getName());
          data.append(UserNotification.CollectionShared.COLLECTION_ICON, resource.getIcon());
          data.append(UserNotification.CollectionShared.COLLECTION_COLOR, resource.getColor());
+
+         final String query = "{\"s\":[{\"c\":\"" + resource.getId() + "\"}]}";
```
kubedo8, line 141: Maybe it will be better to use Query object and customize `toString()` method for this purpose. It's not good idea to hardcode important strings like this.

PR #678, opened 2020-12-31
User notification settings endpoint

kubedo8:


api/src/main/java/io/lumeer/api/model/User.java
```diff
@@ -51,8 +52,7 @@
    public static final String REFERRAL = "referral";
    public static final String AFFILIATE_PARTNER = "affiliatePartner";
    public static final String EMAIL_VERIFIED = "emailVerified";
-   public static final String NOTIFICATIONS = "notifications";
-   public static final String NOTIFICATIONS_LANGUAGE = "notificationsLanguage";
+   public static final String NOTIFICATIONS_SETTINGS = "notifications";
```
marvec, line 55: Why we need to rename the constant? It does not reflect the value now.

PR #805, opened 2021-04-15
Add the ability to clean sample project data

marvec:


lumeer-remote/src/main/java/io/lumeer/remote/rest/ProjectService.java
```diff
@@ -55,6 +55,8 @@
 @Path("organizations/{organizationId:[0-9a-fA-F]{24}}/projects")
 public class ProjectService extends AbstractService {
 
+   public static final String DELETE_SAMPLE_DATA_CONFIRMATION = "PERMANENTLY DELETE";
```
kubedo8, line 58: We require from user to type this in any language?

PR #857, opened 2021-08-11
Reset cron execution on rule change

kubedo8:


api/src/main/java/io/lumeer/api/model/Collection.java
```diff
@@ -273,7 +273,6 @@ public void patch(final Collection resource, final Set<RoleType> roles) {
       }
       if (roles.contains(RoleType.TechConfig)) {
          setPurpose(resource.getPurpose());
-         setRules(resource.getRules());
```
marvec, line 276: I believe this is what breaks tests...

kubedo8: You are right but not in a way you think :)

PR #887, opened 2021-12-19
Get variables and send SMTP blocks

marvec:


lumeer-core/src/main/java/io/lumeer/core/task/executor/bridge/LumeerBridge.java
```diff
@@ -362,6 +370,124 @@ private String getUpdateFileAttachmentsList(final LinkBridge l, final String att
...
+   @SuppressWarnings("unused")
+   public String getVariable(final String variableName) {
+      try {
+         final List<ResourceVariable> vars = task.getDaoContextSnapshot().getResourceVariableDao().getInProject(task.getDaoContextSnapshot().getOrganizationId(), task.getDaoContextSnapshot().getProjectId());
+
+         if (vars != null) {
+            return vars.stream().filter(v -> v.getKey().equals(variableName)).map(ResourceVariable::getValue).map(String::valueOf).findFirst().orElse("");
+         }
+
+         return "";
```
kubedo8, line 382: Please create method to directly load variable from DB. We should avoid draining database whenever it's possible.

kubedo8 (changes requested):

PR #904, opened 2022-01-23
Add user logins info APIs

kubedo8:


lumeer-remote/src/main/java/io/lumeer/remote/rest/UserLoginsService.java
```diff
@@ -0,0 +1,61 @@
...
+
+   @GET
+   @Path("info")
+   public List<OrganizationLoginsInfo> getOrganizationsLoginsInfo(@QueryParam("descending") boolean descending) {
+      return organizationFacade.getOrganizationsLoginsInfo(descending);
+   }
+
+   @GET
+   @Path("delete")
+   public List<OrganizationLoginsInfo> getOrganizationsLoginsInfoToDelete(@QueryParam("descending") boolean descending, @QueryParam("monthsOld") int months) {
```
marvec, line 52: What happens when the descending parameter is not passed? Doesn't it throw an exception? We typically use `Boolean` to allow null values.

api/src/main/java/io/lumeer/api/model/OrganizationLoginsInfo.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package io.lumeer.api.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Set;
+
+public class OrganizationLoginsInfo  {
```
marvec, line 27: The list of templates used in the projects is missing...

marvec (changes requested):

lumeer-remote/src/main/java/io/lumeer/remote/rest/UserLoginsService.java
kubedo8: Default value is `false`, so no exception will be thrown.

api/src/main/java/io/lumeer/api/model/OrganizationLoginsInfo.java
kubedo8: Where can I find information about used template?

PR #993, opened 2023-05-03
Fix project code length

marvec:


kubedo8 (approved):

kubedo8 (changes requested):
We should keep maximum code length to 6 digits

